PR #377, opened 2022-01-28
Delete-AV-detected-xls-files

Shirofune-Security:
Deleted two Excel files as AV was mistakenly alerting on them.


hitenkoku (approved):
Thanks for your pull request. I checked.

PR #379, opened 2022-01-29
Update-readme

Shirofune-Security:
Added WELA link and deleted explanation about using the 0.6.7 library as we are using 0.7.2 by default now.


hitenkoku (approved):
Thanks for your pull request. I checked

PR #581, opened 2022-06-10
change tool chain to stable

hitenkoku:
## What Changed

- changed toolchain rust and cargo which is used in github actions from nightly to stable .


YamatoSecurity (approved):
OKです！

PR #660, opened 2022-08-12
Fix race condition in insert_message. #639

### LINKED ISSUE #639: [bug] v1.4.2+ Hayabusa will randomly miss a few events
**Describe the bug**

Different detection results for each run to same rule and directory

**Step to Reproduce**
1. Download [hayabusa-1.4.2-windows-64-bit.zip](https://github.com/YamatoSecurity/hayabusa/releases/download/v1.4.2/hayabusa-1.4.2-windows-64-bit.zip) from release page and unzip.
2. Open command prompt(Run as Administrator).
3. Execute following command  
   ` hayabusa-1.4.2-win-x64.exe -d ../hayabusa-sample-evtx -o 1.csv`
4.  Execute following command  
   ` hayabusa-1.4.2-win-x64.exe -d ../hayabusa-sample-evtx -o 2.csv`
5. difference detection count 3. result and 4. result.

**Expected behavior**

Same detection results count to same rule and directory.

**Screenshots**

**Environment (please complete the following information):**
 - OS:  Microsoft Windows 11
 - hayabusa version: v1.4.2

**Additional context**

fukusuket:
## What Changed
- Fix #639.
- Fixed static variable `MESSAGES` Read/Write operartion to be thread safe in insert_message.
- Add race condition reproduction unit test.
  - Since [it will fail if not run in a single thread,](https://github.com/Yamato-Security/hayabusa/issues/639#issuecomment-1213229674), add ignore attribute by default.

## Evidence
Run the following command 5 times and confirm that the number of detections is the same.
`hayabusa -d ../hayabusa-sample-evtx -o ...`

Attach the command execution results.
[test-results.zip](https://github.com/Yamato-Security/hayabusa/files/9327948/test-results.zip)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,6 +770,7 @@ dependencies = [
  "pbr",
  "prettytable-rs",
  "quick-xml",
+ "rand",
  "regex",
  "serde",
  "serde_derive",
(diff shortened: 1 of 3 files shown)


YamatoSecurity (approved):
I was able to confirm that the results are stable and the same as version 1.4.1.
Thank you so much for the fix!

hitenkoku (approved):
LGTM

kazuminn:
many thanks.
LGTM.

PR #669, opened 2022-08-20
Timesketch compatibility

### LINKED ISSUE #668: Add output profile for Timesketch
Timesketch needs some mandatory CSV headers to import correctly.

YamatoSecurity:
Closes #668


hitenkoku (approved):
ありがとうございます。問題ないと思います

eljeffeg:
Timesketch doesn't appear to recognize the import  as an `EVTX` for the purposes of running other tools, time  evtx_gap.  Looks like the query is on data_type `query = 'data_type:"windows:evtx:record"'`.  Can this be added to the profile?

https://github.com/google/timesketch/blob/db986692f1f73614290ba9116f0d28d6b1c01062/timesketch/lib/analyzers/win_evtxgap.py#L68

YamatoSecurity:
@eljeffeg It looks like Timesketch is requiring the original .evtx files to parse for this. We only output to CSV and JSON files, we cannot output to .evtx files. So why don't you just upload the original .evtx files for this?

eljeffeg:
I'm confused.  Timesketch doesn't support uploading .evtx files.  It only supports csv and json.  I've recently been working on a fork of Timesketch that does accept evtx files (and zip of evtx files), runs them through hayabusa and processes them.  I haven't seen a way to do evtx directly.

eljeffeg:
I think it might need a column of `type` that is equal to `windows:evtx:record` but I haven't tested that yet.

YamatoSecurity:
@eljeffeg I see, it must be my mistake then. I am pretty busy at the moment so if you could test if it works when a column of type windows:evtx:record is added and you can do the evtx_gap analysis, I would be happy to add it to the default timesketch profile.

eljeffeg:
That didn't seem to work, so not sure how it determines that but would be nice if it did.

PR #745, opened 2022-10-13
fix: profile option command usage.

fukusuket:
## What Changed
- Fix profile option command usage
  - because **verbose-all-field-info**, **verbose-details-and-all-field-info** could not be set.

## Evidence
スクリーンショット 2022-10-14 1 44 20

But above usage is too long ... ? 😅  Is it better to omit ?

### DIFF
--- a/src/detections/configs.rs
+++ b/src/detections/configs.rs
@@ -228,7 +228,7 @@ pub struct Config {
     #[clap(help_heading = Some("FILTERING"), long = "exclude-status", multiple_values = true, value_name = "STATUS")]
     pub exclude_status: Option<Vec<String>>,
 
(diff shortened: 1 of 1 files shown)


YamatoSecurity (changes requested):
Thank you!
Unfortunately it is going to be very long so I'm going to add an issue about printing a list of the different profiles. (I'll probably add even more profiles and change names, etc.. as time goes on) 
For now, can you change this to `Specify output profile (minimal, standard, verbose, all-field-info, etc...)` ?

fukusuket:
Thank you for your prompt review !

>For now, can you change this to Specify output profile (minimal, standard, verbose, all-field-info, etc...) ?

Changed the text above :)
スクリーンショット 2022-10-14 8 46 30

YamatoSecurity (approved):
LGTM!
Thank you so much!

Issue #758, opened 2022-10-18
[bug] Hayabusa cannot update rules when running with Admin privileges


YamatoSecurity:
**Describe the bug**
When hayabusa is running from an Administrator prompt or as SYSTEM, the following error will be displayed when trying to update rules with `-u`:

```
Attempting to git clone the hayabusa-rules repository into the rules folder.
[ERROR] Failed to git clone into the rules folder. Please rename your rules folder name. './rules' exists and is not an empty directory; class=Invalid (3); code=Exists (-4)
[ERROR] Failed to update rules. Error { code: -1, klass: 0, message: "" }
```

**Step to Reproduce**
1. Run command prompt or terminal with administrator. run `hayabusa.exe -u`

Optional: To test with SYSTEM privs: `psexec.exe -i -s cmd.exe`

**Expected behavior**
Hayabusa should be able to update rules regardless of what account run as.

**Environment (please complete the following information):**
 - OS: Win 10
 - hayabusa version: 1.4~1.7.2

fukusuket:
**Step to Reproduce(add a little)**
1.  Log in as a **normal user who is not an administrator** 
2. Run command prompt or terminal with administrator. run hayabusa.exe -u

**The cause of this issue**
Since the following message was output,
`Attempting to git clone the hayabusa-rules repository into the rules folder.`

An error occurs at `Repository::open()` in the code block below.
https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/options/update.rs#L44-L49

So I created a minimal sample that executes `Repository::open()` as follows.
```
use std::path::Path;
use git2::Repository;

fn main() {
    let rule_path = "C:\\Users\\foo\\Desktop\\hayabusa-1.7.2-all-platforms\\rules";
    let hayabusa_rule_repo = Repository::open(Path::new(rule_path));
    match hayabusa_rule_repo {
        Err(err) => panic!("{:?}", err),
        _ => println!("success!")
    };
}
```
and run with `RUST_BACKTRACE=full` then the following stack trace was printed.
```
C:\Users\foo\Desktop\hayabusa-1.7.2-all-platforms>hoge.exe
thread 'main' panicked at 'Error { code: -3, klass: 7, message: "config value 'safe.directory' was not found" }', src\main.rs:10:21
stack backtrace:
   0:     0x7ff794d2070f - <unknown>
   1:     0x7ff794d2ff8a - <unknown>
   2:     0x7ff794d1e9b9 - <unknown>
[...]

PR #769, opened 2022-10-21
fix: improve error msg when update-rules failed with Admin privileges

fukusuket:
## What Changed
- fix: #758 
- improve error msg update-rules failed when executed with permissions different from the permissions of the folder.
  - because [by default, git denies access to the owner's .git repository by a user other than the current user](https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory)
  - In the above situation, [libgit2 return error code -4](https://github.com/libgit2/libgit2/blob/v1.5.0/include/git2/errors.h#L26).

#### Environment 
- Windows Server 2016 Datacenter
- AWS
  - Log in as foo user and execute with Administrator privileges
管理者権限変更後のエラーメッセージ

  - Log in as foo user and execute with foo user
一般ユーザで成功

  - Log in as Administrator and execute
  管理者ユーザで成功

#### Environment 
- Windows 10 Home
- Laptop

  - When not connected to the Internet
ネットに繋がってないとき

  - When there is no rules directory
rulesディレクトリがないとき

  - When there is no .git/config file
gitconfigがないとき


YamatoSecurity (approved):
LGTM!
対応、ありがとうございました！
You need...とYou can...を揃うために空白８個を入れました。
もう少し見栄えが良くなったと思います。

PR #781, opened 2022-10-26
fix: readme linux chmod command example

fukusuket:
**What Changed**

Fixed because the release linux binary file name(hayabusa-1.7.2-**lin**-gnu) and manual (hayabusa-1.7.2-**linux-x64**-gnu)were a bit different.


YamatoSecurity (approved):
OKです。ありがとうございました！

PR #793, opened 2022-11-05
To reduce ram usage and performance, Replaced String with other crate

### LINKED ISSUE #789: Memory usage has almost doubled after introducing custom profiles
With the introduction to custom profiles in 1.5.0, Hayabusa has been using about twice as much memory.
In my tests against a 14GB dataset, with 1.4.3, Hayabusa used less than 6 GB but on the current 1.8.0-dev version, it uses about 11.5GB. We are currently investigating how we can reduce the memory usage.

Note: this only happens when there is enough available memory on the system. When there is low memory, Hayabusa will only use around 300~400MB on average in order not to max out memory.

### LINKED ISSUE #790: Reduce memory usage with Nested, compact_str, etc...
Reduce string memory usage with Nested, compact_str, etc...
https://github.com/Yamato-Security/hayabusa/tree/test-reduce_ram_usage_remove_linked_hashmap 用のissueです。

hitenkoku:
Closes #790 , Closes #789

## What Changed

- To improve memory usage, removed a few call static variables.
- To improve memory usage, replaced Vec<String> with Nested<String>.
- To improve memory usage, replaced String Object in DetectInfo with CompactString.
- Updated changelog.

## Evidence

analyzed result to 2GB evtx file by the following environment.

- CPU: Ryzen 5700G
- RAM: 16GB
- OS: Windows 11

|step| prev memory usage| in 4701a41 memory usage(prev step diff percent)|improved| in fba1688 |
|:---| :---| :---| :---|:--- |
| start process | 31% | 24% | -| 29%|
| finished analyze | 94%(+63%) | 78%(+54%)|  18%| 82%(+53%)|
| finished output | 94%(0%) | 80%(+2%) | 16%  | 83%(+1%)|

output time ( in 4701a41)
```
Elapsed time: 00:03:28.181
Saved file: output.csv (610.9 MB)
```

output time(in fba1688 )

```
Elapsed time: 00:03:26.145
Saved file: output-2.csv (610.9 MB)
```

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -18,7 +18,7 @@
 - プロファイルの`%RecordInformation%` エイリアスを `%AllFieldInfo%` に変更した。 AllFieldInfoフィールドをJSONオブジェクト形式で出力するように変更した。 (#750) (@hitenkoku)
 - AllFieldInfoフィールドのJSONオブジェクト内で利用していたHBFI-プレフィックスを廃止した。 (#791) (@hitenkoku)
 - `--no-summary`  オプションを使用したときに、表示しないルール作者および検知回数の集計を省略した。 (#780) (@hitenkoku)
-- メモリ使用量を少なくし、処理速度を改善した。 (#778) (@hitenkoku)
+- メモリ使用量を少なくし、処理速度を改善した。 (#778 #789 #790) (@hitenkoku)
 
 **バグ修正:**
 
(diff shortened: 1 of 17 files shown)


YamatoSecurity (approved):
LGTM!

hitenkoku:
確認ありがとうございます。マージさせていただきました

fukusuket:
Compared to the main branch before merging, I have confirmed the following performance improvements!!
([6.1GB evtx](https://github.com/NextronSystems/evtx-baseline/releases/download/v0.7/all-evtx.tgz)) 
- before-fix: Mem(avg):6.85 Mem(max):9.48
- after-fix : Mem(avg):6.45 Mem(max):8.89

Mem(avg): -0.40GB Mem(max):-0.59GB Great imporvements ! 🚀
(and I also confirmed that the number of detections was the same before and after fix.)

LGTM:)

Issue #888, opened 2023-01-21
CVE-2023-22742 : SSH MITM vulnerability


YamatoSecurity:
Git2 and libgit2 was not checking SSH certificates so an SSH machine in the middle attack is possible. 
An attacker who could perform a MITM attack could potentially modify rules and config files.

YamatoSecurity:
closed by https://github.com/Yamato-Security/hayabusa/pull/889

PR #889, opened 2023-01-21
#888 fix mitm CVE 2023 22742 vulnerability

YamatoSecurity:
I updated the git2 and libgit2 crates to prevent a SSH MITM attack when updating rules.


hitenkoku (approved):
LGTM

PR #901, opened 2023-02-02
update options

YamatoSecurity:
I updated the following options:
`-t, thread-number` -> `-t, --threads`
`-H, html-report` -> `-H, --HTML-report`
`-e, --eid-filter` -> `-E, --EID-filter`


hitenkoku (approved):
LGTM

PR #909, opened 2023-02-06
update changelog

YamatoSecurity:
I updated the changelog. 
Note: I refer to subcommands as just "commands" because it is easier for non-programmers and that is how we refer to them in the documentation. So I don't want people to be confused if we use both subcommands and commands terms.


hitenkoku (approved):
Thank you for pull request.

LGTM

PR #915, opened 2023-02-10
Fixed  panic of getting file size metadata

hitenkoku:
## What Changed

- Fixed  panic of getting file size metadata

## Evidence

I could not reproduce the reported bug in my environment.


YamatoSecurity (approved):
LGTM! Thank you!

PR #931, opened 2023-02-18
finalize 2.2.2

YamatoSecurity:
If there are no other PRs you two want to add before 2/22 in the 2.2.2 release, then I think it is good as it is now.

Comparison with 2.2.0:
Total & Unique detections: same
Speed: 25:42 -> 26:09 (About 1.7% slower) I think this is because the new evtx crate is slower to avoid parse errors and we shorten channel names generically as well. But I think these two things are worth the slight speed decrease.
Memory: 6.5GB -> 6.4GB (Seems to use slightly less memory)
File size: 1,239,657 KB -> 1,239,654 KB (Only 3KB smaller with smaller channel names...)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -803,7 +803,7 @@ dependencies = [
 
 [[package]]
 name = "hayabusa"
-version = "2.2.2-dev"
+version = "2.2.2"
 dependencies = [
  "aho-corasick",
  "base64 0.21.0",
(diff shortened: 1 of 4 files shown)


fukusuket (approved):
LGTM!🚀

hitenkoku (approved):
LGTM

PR #948, opened 2023-03-01
Added support contains all field

hitenkoku:
## What Changed

- Added support `|contains|all` 

## Evidence

passed in test


YamatoSecurity (approved):
LGTM! I was able to confirm it working with `./target/release/hayabusa json-timeline -d ../hayabusa-sample-evtx -r ../sigma/rules/windows/process_creation/proc_creation_win_cmd_delete.yml`
Also, speed and memory does not seem to be affected.

PR #961, opened 2023-03-15
feat: Support `cidr` rule

### LINKED ISSUE #955: Support `|cidr`
Some rules are now using `|cidr`.
Example (builtin/security/win_security_successful_external_remote_rdp_login.yml)
```
title: External Remote RDP Logon from Public IP
id: 259a9cdf-c4dd-4fa2-b243-2269e5ab18a2
detection:
    selection:
        EventID: 4624
        LogonType: 10
    filter:
        SubjectUserName: '-'
        IpAddress|cidr:
            - 10.0.0.0/8
            - 172.16.0.0/12
            - 192.168.0.0/16
            - 224.0.0.0/4
            - 127.0.0.0/8
    condition: selection and not filter
```

I think we can use the `cidr-utils` crate to handle this. 
@fukusuket Are you interested in this issue?

fukusuket:
## What Changed

- Fixed #955 

### Environment 
- OS: macOS montery version 13.1
- Hard: Macbook Air(M1, 2020) , Memory 8GB, Core 8

### Test

`test.yml` is as follows.(no cidr match)
```test.yml
detection:
    selection:
        Channel: Security
        EventID: 4624
    condition: selection 
```

then SrcIP uniq count is as follows
`./hayabuse json-timeline -d ./hayabusa-sample-evtx -r test.yml -o result.json` 
`cat result.json | jq .Details.SrcIP | sort | uniq -c`
```
 518 "-"
  10 "10.0.2.17"
  63 "10.23.123.11"
  99 "10.23.23.9"
   1 "10.23.42.18"
   5 "10.23.42.22"
   1 "10.23.42.30"
   6 "10.23.42.38"
  67 "127.0.0.1"
   3 "172.16.66.142"
   4 "172.16.66.19"
   2 "172.16.66.25"
   5 "172.16.66.37"
   1 "192.168.1.100"
   2 "192.168.1.2"
   1 "192.168.1.200"
   4 "192.168.1.219"
   1 "192.168.111.87"
   1 "192.168.20.11"
  38 "::1"
   4 "fe80::1cae:5aa4:9d8d:106a"
   6 "fe80::7191:d555:270f:4d0b"
   4 "fe80::79bf:8ee2:433c:2567"
   5 "fe80::e50e:b89e:4718:3aa"
```

`test.yml` is as follows.(with cidr match)
```
detection:
    selection:
        Channel: Security
        EventID: 4624
    filter:
        IpAddress|cidr: 10.0.0.0/8
    condition: selection and filter
```

then SrcIP uniq count is as follows
`./hayabuse json-timeline -d ./hayabusa-sample-evtx -r test.yml -o result.json` 
`cat result.json | jq .Details.SrcIP | sort | uniq -c`
```
  10 "10.0.2.17"
  63 "10.23.123.11"
  99 "10.23.23.9"
   1 "10.23.42.18"
   5 "10.23.42.22"
   1 "10.23.42.30"
   6 "10.23.42.38"
```
I confirm that the number of `SrcIP` matches before and after the `cidr` filter


YamatoSecurity:
@fukusuket Thanks so much!
Looks good. Just one bug. When the `IpAddress` field does not exist, hayabusa crashes.
Test rule:
```
title: Test CIDR
id: 259a9cdf-c4dd-4fa2-b243-2269e5ab18a4
status: experimental
description: 
references:
author: hoge
date: 2023/01/19
tags:
logsource:
    product: windows
    service: security
detection:
    selection:
        IpAddress|cidr:
            - 10.0.0.0/8
            - 172.16.0.0/12
            - 192.168.0.0/16
            - 224.0.0.0/4
            - 127.0.0.0/8
    condition: selection
level: medium
```
When I run `./target/release/hayabusa csv-timeline -d ../hayabusa-sample-evtx -r ../hayabusa-rules/hayabusa/test.yml` I get the following error:
```
thread 'hayabusa-thread' panicked at 'called `Option::unwrap()` on a `None` value', src/detections/rule/matchers.rs:507:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(5017), ...)', src/detections/detection.rs:167:41
```
If I change `IpAddress|cidr` to `IpAddress`, `IpAddress|contains`, etc.. it does not crash.
Could you check this?

src/detections/rule/matchers.rs
```diff
@@ -502,10 +497,26 @@ impl LeafMatcher for DefaultMatcher {
...
-        if pipe.get_eqfield().is_some() {
-            return pipe.is_eqfield_match(event_value, recinfo);
+        let pipe: &PipeElement = self.pipes.first().unwrap_or(&PipeElement::Wildcard);
+        let match_result = match pipe {
+            PipeElement::EqualsField(_) | PipeElement::Endswithfield(_) => {
+                Some(pipe.is_eqfield_match(event_value, recinfo))
+            }
+            PipeElement::Cidr(ip_result) => match ip_result {
+                Ok(matcher_ip) => {
+                    let event_value_str = event_value.unwrap();
```
hitenkoku, line 507: It crashes here when the IpAddress field does not exist, so you might want to modify it as follows

```suggestion
                    let event_value_str = event_value.unwrap_or_default();
```

hitenkoku (changes requested):
@fukusuket Thank you for the pull request.
Please check the following points that have been brought to our attention.

src/detections/rule/matchers.rs
hitenkoku: This fix should solve the problem with https://github.com/Yamato-Security/hayabusa/pull/961#issuecomment-1470976731

src/detections/rule/matchers.rs
```diff
@@ -2395,4 +2408,80 @@ mod tests {
 
         check_select(rule_str, record_json_str, false);
     }
+
+    #[test]
```
hitenkoku, line 2412: Can you please create a test for a test case where there is no IpAddress|cidr in the rule, but there is no IpAddress field in the record?
(This fix should solve the problem with https://github.com/Yamato-Security/hayabusa/pull/961#issuecomment-147097673 test case. The expectation in that test case is that the number of cases to be detected is zero.)

hitenkoku (commented):
Please check the comments added about the test cases.

src/detections/rule/matchers.rs
fukusuket: Thank you so much for looking into the cause of the problem! :)

src/detections/rule/matchers.rs
fukusuket: I will also add test case and check it!

YamatoSecurity (approved):
LGTM! Thank you!

fukusuket:
Thank you so much for quick review and test :) 
@hitenkoku  I've fixed the parts you reviewed. I would appreciate it if you could check🙇

hitenkoku (approved):
Thanks.LGTM.

PR #962, opened 2023-03-16
finalize 2.3.0

YamatoSecurity:
I want to release 2.3.0 earlier so that we can test it before the TMCIT event.
Please check.


fukusuket (approved):
LGTM!🚀

PR #966, opened 2023-03-17
Fixed bug update rules command panics on 2.3.0

hitenkoku:
## What Changed

- Fixed panics error when `update-rules` subcommand execute on version 2.3.0.

## Evidence

I confirmed that the update-rules subcommand can be executed on the modified hayabusa without causing panic as follows

```
 >./963.exe update-rules -q
...
Start time: 2023/03/17 22:49

 - Autorun Keys Modification (Modified: 2022/05/14 | Path: rules\sigma\builtin\deprecated\registry_event_asep_reg_keys_modification.yml)
 ...
 - Disabled Users Failing To Authenticate From Source Using Kerberos (Modified: 2023/03/13 | Path: rules\sigma\builtin\unsupported\win_security_susp_failed_logons_single_source_kerberos2.yml)

Updated Sigma rules: 205
Rules updated successfully.

There is a new version of Hayabusa: v2.3.0
You can download it at https://github.com/Yamato-Security/hayabusa/releases

```

I would appreciate it if you could review.


YamatoSecurity (approved):
LGTM! Thank you so much!

PR #968, opened 2023-03-18
add short options

YamatoSecurity:
The `--enable-deprecated-rules` and `--enable-unsupport-rules` are useful options that I want to use frequently but are long to type so I added short options. Please check if this is ok.


hitenkoku (approved):
Thanks for your pull request.
I checked source code and corfirm binary(csv-timeline subcommand)

LGTM.

PR #988, opened 2023-04-05
perf: replaced trim method in afterfact with process in detection

hitenkoku:
## What Changed

- replaced trim method in afterfact with process in detection


YamatoSecurity (approved):
LGTM! unfortunately the speed doesn't seem to change that much but seems to use about 100MB less memory in my tests. So let's use this if it seems ok.
@fukusuket Could you check the source code before we merge it?

fukusuket (approved):
I checked source code and Integration Test passed :) LGTM!!🚀
https://github.com/Yamato-Security/hayabusa/actions/runs/4627367137

Issue #989, opened 2023-04-06
Part of hayabusa help menu is green in 2.3.3-dev


YamatoSecurity:
I think due to a change in the most recent version of clap, the message and usage are now printed in green because the logo is outputted in green.
I tested this happens on mac and also Windows Terminal.
<img width="730" alt="Screen Shot 2023-04-06 at 9 31 53 PM" src="https://user-images.githubusercontent.com/71482215/230379493-48792b77-edd1-4030-944b-f68f48330776.png">

@hitenkoku Could you take a look at this?

PR #990, opened 2023-04-06
UI(main): fixed part of help menu green color #989

hitenkoku:
## What Changed

- replaced message and usage color green with white

## Evidence

image


YamatoSecurity (approved):
LGTM! There were just two other places that needed to updated as well after the logo gets printed but was easy to do so I updated those on my end. Thank you!

Issue #994, opened 2023-04-10
Remove action-rs in GitHub CI


hitenkoku:
The action-rs used by Hayabusa's GithubCI has issued a deprecation notice and has stopped being updated.
Therefore, it would be better to modify the action-rs in Hayabusa so that it is not used.

refs(Japanese): https://techblog.paild.co.jp/entry/2023/04/10/170218

PR #995, opened 2023-04-10
ci(rust/coverage): removed action-rs #994

hitenkoku:
## What Changed

- removed action-rs from CI


hitenkoku:
@fukusuket @YamatoSecurity 
Fixed to not use action-rs in CI within Hayabusa. Would you check that it is working without any problem?

YamatoSecurity (approved):
LGTM! Thank you!

PR #1005, opened 2023-04-13
finalize 2.4.0

YamatoSecurity:
I will be busy in Korea from tomorrow so I would like to finalize 2.4.0 as it is now.
Could you please check?

By the way, I changed "Event Frequency Timeline" to "Detection Frequency Timeline" because the timeline is of what was detected and not the original events (including events w/o detections). 
"Event Frequency Timeline" could be confused as showing all of the original events.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,42 +39,51 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.2.6"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


fukusuket (approved):
I have verified that the integration tests pass:) LGTM!🚀
https://github.com/Yamato-Security/hayabusa/actions/runs/4694768187

hitenkoku (approved):
Thank you for your response!

Sorry for the late review!

LGTM

YamatoSecurity:
@fukusuket @hitenkoku Thank you for checking!

PR #1014, opened 2023-04-18
updated aho corasick crate

hitenkoku:
## What Changed

- updated Aho-Corasick crate to version 1.0


YamatoSecurity (approved):
I checked it and all looks good.
Thanks so much for the update!

PR #1107, opened 2023-06-26
perf: changed from regex match to split() when parse output result

fukusuket:
## What Changed

- Changed from `regex match` to [str::split](https://doc.rust-lang.org/std/primitive.str.html#method.split) when parse output result.
  -  [Because String standard methods are faster than regex in many cases](https://github.com/Yamato-Security/hayabusa/blob/main/doc/RustPerformance-English.md#use-standard-string-methods-instead-of-regular-expressions)

Since this improvement effect is small, I think that the speed will improve proportionally as the number of detections increases.
(I could see an improvement on my `Windows`, but no change on my `M1 Mac`😅)

### Environment
- OS: Windows 10 Home edition
- Hard: Memory 16GB , Core 8, SSD, laptop

### Test(all-evtx.tgz(6.1GB))
I got the benchmark 3 times on Windows, and I confirmed that all 3 times it improved to the almost same extent as the following.
|                    | This PR| main |
|-------------------|----------------|----------------|
| **Elapsed time** | **00:11:16.687** | **00:11:32.533** |
| Rule Parse Processing Time | 00:00:27.645| 00:00:27.251|
| **Analysis Processing Time**| **00:10:25.069**|**00:10:39.605**|
| Output Processing Time | 00:00:23.968| 00:00:25.667|
| Memory usage | 6.0 GiB| 6.0 GiB|
| Total Detections  | 1,627,313 |1,627,313 |
| Unique Detections | 158 |158| 
| File Size         | 1352703683 bytes |1352703683 bytes  | 

However, there was no improvement effect on `M1 Mac`, so the effect may differ depending on the OS.
(I confirmed no slowdown on `M1 Mac`)


hitenkoku (approved):
Thanks for your pull request.
LGTM

YamatoSecurity:
I took benchmarks:

|                    | This PR| main |
|-------------------|----------------|----------------|
| **Elapsed time** | **40:24** | **39:34** |
| Rule Parse Processing Time | 00:10 | 00:17 |
| **Analysis Processing Time**| **36:54**| **36:02**|
| Output Processing Time | 3:19 | 3:14 |
| Memory usage | **42.0 GiB** | **41.5 GiB** |
| Total Detections  | 16,317,841 | 16,317,841 |
| Unique Detections | 665 |665| 
| File Size         | 10.3 GB | 10.3 GB  | 

LGTM! It is indeed faster and uses 500MB less memory. Thank you!

fukusuket:
@YamatoSecurity 
Thank you so much for taking the benchmark :) Are the above results of `This PR` and `main` reversed?

YamatoSecurity:
@fukusuket Oh no! Actually they aren't reversed...(T_T) It looks like this PR was slower in my tests... let me run the benchmarks again.

fukusuket:
@YamatoSecurity 
I have a question! Did you reboot the OS before acquiring the benchmark?
(This is because the speed changes depending on whether the evtx file is on the memory cache)

I see... If the next benchmark shows the same result, I'm sorry but I'll close this PR :(
Thank you for taking the benchmark again🙇

YamatoSecurity:
@fukusuket Thanks for reminding me! No, I did not reboot the OS which may be the cause. I will try to do the benchmarks better this time. =)

fukusuket:
@YamatoSecurity 
I see! there may also be a file memory cache effect... 🤔

I hope the following information is helpful.
https://github.com/Yamato-Security/hayabusa/blob/main/doc/RustPerformance-English.md#use-windows-performance-counter

I've also recently noticed that my M1 Mac can get hot and slow under constant high CPU load.
This makes it more difficult to get accurate benchmarks...😢

YamatoSecurity:
@fukusuket I took benchmarks on two more different sets of data and rebooted in between but unfortunately, all times the main branch was faster and used a little less memory:
```
20Gb logs:
this PR:
elasped time: 40:49.5
rule parse: 10.5
analysis: 37:18
output 3:20
memory: 42 GB

main:
elasped time: 40:40
rule parse: 15.9
analysis: 37:11
output 3:13
memory: 41.5 GB


13.7GB logs:
this PR:
elasped time: 30:41.8
rule parse: 00:00:04
analysis: 29:20
output 1:16
memory: 10.3 GB

main:
elasped time: 29:59
rule parse: 00:04
analysis: 28:26
output 1:27
memory: 10.2 GB
```

Maybe the regex crate is now more efficient than split? I am not sure. Benchmarks seem to differ by environment and dataset but I want to optimize for large datasets on a Windows environment so unfortunately it might be better to keep things as they are.
Thank you so much for the PR though!

fukusuket:
@YamatoSecurity Thank you so much for taking the benchmark🙇
I see, split may be slower... It seems that there are more cases where the speed slows down, so I will close this PR!

PR #1122, opened 2023-07-03
fixed result summary table breaks

hitenkoku:
## What Changed

- fixed result summary table breaks

## Evidence

image


YamatoSecurity (approved):
Now the table doesn't break even with many events. Thank you!

PR #1129, opened 2023-07-13
renamed `metrics` command `eid-metrics`

hitenkoku:
## What Changed

- renamed `metrics` command `eid-metrics`


YamatoSecurity (approved):
@hitenkoku Thank you for such a fast PR!
LGTM but @fukusuket could you double-check the source code just in case there are any other variables that should be marked as eid-metrics for clarity, etc..

fukusuket (approved):
LGTM!!🚀

PR #1131, opened 2023-07-14
added `include-eid` and `exclude-eid` filtering option in `csv-timeline`,  `json-timeline`, `pivot-keywords-list`

hitenkoku:
## What Changed

- added `include-eid` and `exclude-eid` filtering option in `csv-timeline`,  `json-timeline`, `pivot-keywords-list`


fukusuket (approved):
I confirmed as follows, LGTM!!🚀
- No unnecessary diffs compared to v2.6.0
- No unnecessary diffs compared to After merging the results of --include-eid and --exclude-eid
- No unnecessary diffs compared to jq filtered result and pivot result

YamatoSecurity (approved):
@hitenkoku LGTM! Thanks so much for the implementation!
@fukusuket Thank you so much for checking!

PR #1136, opened 2023-07-20
added `-R --remove-duplicate-data` support in `json-timeline` option

hitenkoku:
## What Changed

- added `-R --remove-duplicate-data` support in `json-timeline` option


YamatoSecurity (approved):
@hitenkoku LGTM!
I tested this on 128GB of data. 
Without de-duplication it finishes in `5:14:21` with an output file size of 32.1 GB.
With `-R`, it finishes in `5:13:08` with a file size of 27.9 GB. So is actually faster!

fukusuket (approved):
I confirmed that the duplicate data was converted to DUP! LGTM!!🚀

PR #1137, opened 2023-07-21
changed progress bar width

hitenkoku:
## What Changed

- Changed progress bar width to 80% to leave room for adjustment of the terminal.


fukusuket (approved):
I confirmed that the size will be 80% in the following environment! LGTM!!🚀
- macOS 
  - iTerm 
- Windows10 
   - Command rompt
   - PowerShell 7
   - Windows Terminal

YamatoSecurity:
@fukusuket Thank you for checking!
@hitenkoku Sorry for the late reply. I was doing lots of testing. I decided that it looks better with a fixed width. Something similar to the cargo build progress bar. I also added a max refresh rate so that the progress bar doesn't use too much CPU when going fast.
I would also like to show ticks to show that hayabusa is still working even if it is working on a large single evtx file.
```
        pb.show_tick = true;
        pb.tick_format("├┬┤┴");
```
However, we need to call `pb.tick();` from a different function that is scanning the evtx data so that the tick is always moving. (Right now, it only updates the tick when hayabusa moves to a new evtx file which defeats the purpose)
Is this easy to do? If it is difficult, then we can leave it as it is.

YamatoSecurity:
Also, do either of you know if it is possible to show the evtx file path of the currently scanned evtx? Either before or after the progress bar is fine.
Example:
```
Currently scanning: "../hayabusa-sample-evtx/DeepBlueCLI/many-events-application.evtx"

32 / 585 [=>                              ] 5.47 % 34s
```

or something similar to the cargo build progress bar on one line:
```
32 / 585 [=>                              ] 5.47 % 34s  (../hayabusa-sample-evtx/DeepBlueCLI/many-events-application.evtx)
```
Although this might cause line wrapping with long file names..

The tricky thing is I want it to update in the same place, not add new lines like with `--verbose`

hitenkoku:
@YamatoSecurity Changed from pbr to indicatif in 6d272ba11ab3ca5fc1433123e2959bd6641c2bd2.
The reasons are as follows
Because there are many unfulfilled pull-requests in pbr, and the speed of version update is slowing down.
It is difficult to implement tick processing within the framework of pbr.

I have implemented what you have pointed out.
Would you confirm this?

YamatoSecurity:
@hitenkoku Sorry, is it possible to display the evtx file even when `-v` is not specified?
I tried this:
```
        let template = if stored_static.verbose_flag {
            "{human_pos} / {human_len} {spinner} [{bar:40.green}] {percent}% {eta} \r\n\r\n{msg}"
        } else {
            "{human_pos} / {human_len} {spinner} [{bar:40.green}] {percent}% {eta} \r\n\r\n{msg}"
        };
```
but nothing gets displayed when I don't use `-v`.

hitenkoku:
@YamatoSecurity Sorry for the delay in getting back to you.
Fixed to output file name without verbose flag.
Sorry to keep repeating this, but would you confirm this?

hitenkoku:
@YamatoSecurity We apologize.
My correction was incorrect.

https://github.com/Yamato-Security/hayabusa/pull/1137/commits/d9ec86ac545a8f41d48b7b402b0040af52d1ed5a で修正された内容で問題ありません。
The correction made at https://github.com/Yamato-Security/hayabusa/pull/1137/commits/d9ec86ac545a8f41d48b7b402b0040af52d1ed5a is fine.

YamatoSecurity (approved):
@hitenkoku Thank you so much for checking my commit! I tested out the new progress bar on the traditional Command Prompt, mac, and linux and all looks good so I will merge this.

PR #1138, opened 2023-07-21
supported different time formats in the `search` command

hitenkoku:
## What Changed

- Added support of output time format options in `search` command. (`--European-time`, `--ISO-8601`, `--RFC-2822`, `--RFC-3339`, `--US-time`, `--US-military-time`, `-U, --UTC`) 


YamatoSecurity (approved):
@hitenkoku LGTM! Thanks for such a speedy PR!
I checked that all of the times are correct. There is one thing not related to this PR though. `-l, --live_analysis` uses `_` instead of `-`. Can you change it to `-l, --live-analysis`?
(I just noticed it also uses `_` for csv-timeline and all the other commands as well, but since we always use hyphens it is probably better to change to `--live-analysis`)

fukusuket (approved):
I confirmed that all time formats are output correctly! LGTM!!🚀

hitenkoku:
@YamatoSecurity  @fukusuket 
Thanks for your very fast review. I will merge it.

PR #1141, opened 2023-07-23
print rule count by rule id instead of path

hitenkoku:
## What Changed

- Changed rule count by rule ID instead of path. 


fukusuket (approved):
I confirmed that it is counted by ruleID! LGTM! ! 🚀

YamatoSecurity (approved):
@hitenkoku Thanks so much! I changed to the default ticker because it looks better.

YamatoSecurity:
@hitenkoku 
Thanks so much!
```
Deprecated rules: 169 (6.94%) (Disabled)
Experimental rules: 1289 (52.91%)
Stable rules: 196 (8.05%)
Test rules: 951 (39.04%)
Unsupported rules: 45 (1.85%) (Disabled)

Hayabusa rules: 159
Sigma rules: 2277
Total enabled detection rules: 2436
```
This looks good.

```
Excluded rules: 1
Noisy rules: 1 (Disabled)
```
There is only 1 rules for both excluded rules and noisy rules, however, I think there should be more. Can you check this?

hitenkoku:
@YamatoSecurity  

Thank you for your comment.
The process for exclude rule etc. is already in
We have confirmed that they are filtered by unique IDs.
I fixed in https://github.com/Yamato-Security/hayabusa/pull/1141/commits/43aca388d6359fbc0a194a2d1ee2ed8e69474d 83 to correct the problem.

Would you check it out?

> ```
> Excluded rules: 1
> Noisy rules: 1 (Disabled)
> ```
> 
> There is only 1 rules for both excluded rules and noisy rules, however, I think there should be more. Can you check this?

YamatoSecurity (approved):
@hitenkoku Thanks so much! All looks good now so I will merge this.

PR #1151, opened 2023-07-31
1150 timeline start and timeline end do not work with pivot keywords list

hitenkoku:
## What Changed

- Fixed `--timeline-start` and `--timeline-end` do not work correctly with `pivot-keywords-list` command.


YamatoSecurity (approved):
@hitenkoku Thank you so much for fixing this! I confirmed that it works now.

PR #1156, opened 2023-08-08
update readme verbose section

YamatoSecurity:
I updated the verbose output section to the readme because it is different with the latest version.
I also updated the error messages to add spaces to make easier to read and updated crates so now `krapslog` does not use `atty`.


fukusuket (approved):
I confirmed that the integration test passed as follows. LGTM!🚀
https://github.com/Yamato-Security/hayabusa/actions/runs/5802548680/job/15729109627

hitenkoku (approved):
@YamatoSecurity thanks for your pull request.
lgtm.

YamatoSecurity:
@hitenkoku @fukusuket Thank you both for checking! Then I will merge this.

PR #1169, opened 2023-08-30
Remove deprecated chrono function use

hitenkoku:
## What Changed

- Cargo update
- Remove deprecated chrono function use


YamatoSecurity (approved):
@hitenkoku LGTM! thank you!

PR #1205, opened 2023-10-30
add explanation about rule sets

YamatoSecurity:
I added explanations about the rule sets. @fukusuket Could I ask you to do the Japanese translations? I do not have time at the moment and you will probably do a better job than me.


fukusuket (approved):
@YamatoSecurity 
I a dded Japanese README:) I would appreciate it if you could check it out🙏

YamatoSecurity:

PR #1233, opened 2023-12-18
added stricter checks for options

### LINKED ISSUE #1232: Add stricter required options
The following options should only be run without the wizard so I want to display an error message when the usage is incorrect:
```
-D, --enable-deprecated-rules
-n, --enable-noisy-rules
-u, --enable-unsupported-rules
-e, --exact-level
-m, --min-level
      --exclude-category <CATEGORY...>  Do not load rules with specified logsource categories (ex: process_creation,pipe_created)
      --exclude-status <STATUS...>      Do not load rules according to status (ex: experimental) (ex: stable,test)
      --exclude-tag <TAG...>            Do not load rules with specific tags (ex: sysmon)
      --include-tag <TAG...>            Only load rules with specific tags (ex: attack.execution,attack.discovery)
-r, --rules
```

YamatoSecurity:
closes #1232 

I made the required options a little stricter and updated `csv` to 1.3 as it is just a minor update.
Please check when you have time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -462,9 +462,9 @@ dependencies = [
 
 [[package]]
 name = "csv"
-version = "1.2.2"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "626ae34994d3d8d668f4269922248239db4ae42d538b14c398b74a52208e8086"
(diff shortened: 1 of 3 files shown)


hitenkoku (approved):
Thanks for your pull request.
LGTM

fukusuket (approved):
LGTM!!🚀

PR #1237, opened 2023-12-19
Added message about reporting issues and FPs

hitenkoku:
## What Changed

- Added message about reporting issues and FPs when `csv-timeline` or `json-timeline` command was used.


YamatoSecurity (approved):
@hitenkoku LGTM! Thank you!

Issue #787, opened 2022-11-03
Reduce usage of clone()


YamatoSecurity:
We would like to reduce the number of clone() uses to reduce memory usage as much as possible.

@fukusuket would you like to try to fix these? I can assign you. If it is difficult then we can ask someone else to take a look at it.

<img width="1090" alt="clone" src="https://user-images.githubusercontent.com/71482215/199856686-e2c8a097-651e-4d57-9261-4e932f2a3988.png">

fukusuket:
@YamatoSecurity 
I will be happy to fix it :) 
If it is difficult for me to fix it ..., I will ask you to change the assignment !

YamatoSecurity:
Ok, thanks! Do what you can and if there are some difficult ones to fix we can always ask others for help.

fukusuket:
After grepping, the clone() in the v1.7.2 is as follows.
- detections.rule  (10 usages found)
  - condition_parser.rs  (6 usages found)
    - [L75 ConditionToken::SelectionReference(name.clone())](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L75)
    - [L83 ConditionToken::ParenthesisContainer(sub_tokens) => sub_tokens.clone()](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L83)
    - [L84 ConditionToken::AndContainer(sub_tokens) => sub_tokens.clone()](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L84)
    - [L85 ConditionToken::OrContainer(sub_tokens) => sub_tokens.clone()](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L85)
    - [L86 ConditionToken::NotContainer(sub_tokens) => sub_tokens.clone()](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L86)
    - [L87 ConditionToken::OperandContainer(sub_tokens) => sub_tokens.clone()](https://github.com/Yamato-Security/hayabusa/blob/v1.7.2/src/detections/rule/condition_parser.rs#L87)
  - mod.rs  (3 usages found)
[...]

PR #1268, opened 2024-02-03
Remove replace subtoken

hach1yon:
#787 

## What Changed
change parse condition logic in order to remove unnecessary function and in order not to create unnecessary token instance.

## Evidence

The testcases have passed.


fukusuket (approved):
I confirmed that there is no difference with 2.12.0 using the csv-timeline/json-timeline command :) LGTM!!🚀

PR #1270, opened 2024-02-04
reduce the number of AndContainer's and OrContainer's instance

hach1yon:
## What Changed

I reduced the number of AndContainer and OrContainer instance that Hayabusa creates when parsing condition section of rule files. 
* The previous logic creates unnecessary AndContainer. For example, if the condition section is `section1 and section2 and secion3 and section4`, previous logic creates three AndContainer like `AndContainer.(section1, AndContainer.(section2, AndContainer.(section3, secion4)))`. Current logic create only 1 AndContainer like `AndContainer.(section1, section2, section3, section4)` if the condition section is `section1 and section2 and secion3 and section4`.
* The previous logic creates OrContainer even if there is only one subtoken. For example, if the condition section is `section1 `, the previous logic creates OrContainer like `OrContainer(section1)`. The current logic doesn't create unnecessary OrContainer instance.

## Evidence

Please describe the functionality gained by merging this pull-request and the evidence that the bug has been fixed.

### DIFF
--- a/src/detections/rule/condition_parser.rs
+++ b/src/detections/rule/condition_parser.rs
@@ -305,20 +305,27 @@ impl ConditionCompiler {
         }
 
         // 先にANDでつながっている部分を全部まとめる
-        let mut operant_ite = operand_list.into_iter();
-        let mut operands = vec![operant_ite.next().unwrap()];
+        let mut operand_ite = operand_list.into_iter();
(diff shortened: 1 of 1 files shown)


hitenkoku (approved):
Thanks for your pull requet.
LGTM.

I confirmed no differ in csv-timeline output with main branch.

PR #1292, opened 2024-03-03
low memoption wave2

hach1yon:
## What Changed
I have implemented low memory mode.
- You can use low memory mode if you set true to is_low_memory.
    - `stored_static.is_low_memory = true`

## remained
* I haven't added low memory mode to command line option so we have to do it in another PR.
* The following features are disabled in low memory option.
   * sort timeline log
   * remove duplicate log 
       * As well as sorting timeline log, it's difficult to remove duplicate log. This is because it's necessary to load all the timeline logs in order to remove dupplicate log.
* I changed analysis_file() in main.rs in order to supress warning but we can do it better.

## Evidence

Please describe the functionality gained by merging this pull-request and the evidence that the bug has been fixed.

### DIFF
--- a/src/afterfact.rs
+++ b/src/afterfact.rs
@@ -406,9 +406,7 @@ fn init_writer(stored_static: &StoredStatic) -> AfterfactWriter {
             .quote_style(QuoteStyle::Never)
             .from_writer(target),
         Action::CsvTimeline(_) => WriterBuilder::new()
-            .delimiter(b'\n')
-            .double_quote(false)
-            .quote_style(QuoteStyle::Never)
+            .quote_style(QuoteStyle::NonNumeric)
             .from_writer(target),
         _ => panic!("unknown action is specified!!"),
     };
--- a/src/afterfact.rs
+++ b/src/afterfact.rs
@@ -72,6 +72,13 @@ pub struct AfterfactInfo {
     pub rule_title_path_map: HashMap<CompactString, CompactString>,
     pub rule_author_counter: HashMap<CompactString, i128>,
     pub timestamps: Vec<i64>,
+    pub has_displayed_header: bool,
+    pub detected_rule_files: HashSet<CompactString>,
+    pub detected_rule_ids: HashSet<CompactString>,
+    pub detected_computer_and_rule_names: HashSet<CompactString>,
+    pub author_list_cache: HashMap<CompactString, Nested<String>>,
+    pub prev_message: HashMap<CompactString, Profile>,
+    pub prev_details_convert_map: HashMap<CompactString, Vec<CompactString>>,
 }
 
 struct InitLevelMapResult(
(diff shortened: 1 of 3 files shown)


hach1yon:
@YamatoSecurity 
Could you check low memory option with very big event log?
You can enable low memory option if you uncomment the following line.
https://github.com/Yamato-Security/hayabusa/pull/1292/commits/d98e4903d8f2404cdea2250b11ffc2134d88b939#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR81

YamatoSecurity:
@hach1yon Thanks so much! I will do some testing on 130 GB logs. It will take some time so please give me 2-3 days to test.

fukusuket (approved):
I confirmed the following improvements using baseline-evtx data!
It's a great improvement!! LGTM!!🚀 

This PR
```
Elapsed time: 00:10:53.72
...
Rule Parse Processing Time: 00:00:01.870
Analysis Processing Time: 00:10:51.686
Output Processing Time: 00:00:00.114
...
heap stats:     peak       total       freed     current        unit       count
  reserved:     2.0 GiB     2.0 GiB     0           2.0 GiB
```

2.13.0
```
Elapsed time: 00:10:42.375
...
Rule Parse Processing Time: 00:00:01.859
Analysis Processing Time: 00:09:15.389
Output Processing Time: 00:01:25.121
...
heap stats:     peak       total       freed     current        unit       count
  reserved:    20.0 GiB    20.0 GiB     0          20.0 GiB
```
(In addition, I confirmed that there were no differences other than the count conditions:))

YamatoSecurity:
@hach1yon Is it possible to still support count rules with low memory mode? or do you think that it would significantly increase memory usage? If possible it would still be nice to support count rules in low memory mode. (Maybe add another option to enable count rules?)

hach1yon:
> @hach1yon Is it possible to still support count rules with low memory mode? or do you think that it would significantly increase memory usage? If possible it would still be nice to support count rules in low memory mode. (Maybe add another option to enable count rules?)

@YamatoSecurity 
Oh, I'm sorry that I don't know count rules feature. Is it existing feature or new feature?

YamatoSecurity:
@hach1yon The count rules are the ones with `| count` ( https://github.com/Yamato-Security/hayabusa-rules?tab=readme-ov-file#aggregation-conditions-count-rules ) You are the one who implemented this feature a long time ago. :smile:
Most of the count rules are marked as unsupported or deprecated so you can test the difference in detections with:
`./target/release/hayabusa csv-timeline -d ../hayabusa-sample-evtx  -w --debug -u -D -C -o test.csv`
against the [hayabusa-sample-evtx](https://github.com/Yamato-Security/hayabusa-sample-evtx) repository.
When I tested **without** low-memory-mode it found `645` unique detections, but **with** low-memory-mode `636` unique detections so it would be nice to still be able to detect the same amount.

YamatoSecurity:
@hach1yon 
If somebody puts in a very large timeframe for the count rule and there are many logs then I think there might be a noticeable increase in memory but most rules only look for logs in a 5 minute timespan, etc.. so I don't think it should consume too much memory as long as memory is freed after the timespan is over and there is no rule match.

hitenkoku (approved):
LGTM

hach1yon:
@YamatoSecurity 
I'll check it.

hach1yon:
@YamatoSecurity 
I fixed codes so that count rule logs are showed in low memory option.  Could you check?

YamatoSecurity (approved):
@hach1yon LGTM! I tested 130GB of data and it does the same detection as 2.13.0.
Memory usage went down from 106 GB to 5.3 GB and processing time decreased from 5 hours 12 minutes to 4 hours 45 minutes so both are really great improvements. Thank you for this!

PR #1313, opened 2024-03-16
update logo

YamatoSecurity:
I updated the logo to single pipes because they display cleaner on certain monitors.


hitenkoku (approved):
LGTM

PR #1326, opened 2024-04-13
Added support `windash` pipe modifier

hitenkoku:
## What Changed

- Added for support for `windash` keyword.

## Evidence

passed added test for `contains|windash` and `contains|all|windash` modifier.

I would appreciate it if you could check it out when you have time.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -4,7 +4,7 @@
 
 **改善:**
 
-- Sigmaルールの `windash` (`|contains|windash`と`contains|all|windash`)に対応した。 (#1319) (@hitenkoku)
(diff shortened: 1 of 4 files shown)


fukusuket (approved):
I confirmed following two points :) LGTM!!🚀
- There is no difference in the results of csv-timeline .. -D -n -u -w
- [win_system_susp_service_installation_script.yml](https://github.com/SigmaHQ/sigma/blob/93b71ec1bbcff325709b07f0db42947690d7dee8/rules/windows/builtin/system/service_control_manager/win_system_susp_service_installation_script.yml#L4) 's detection
  - There is no difference in detection results after converting to logic without windash.

YamatoSecurity (approved):
@hitenkoku Thanks so much! LGTM! I checked that both `contains|windash` and `contains|all|windash` are working properly.

PR #1331, opened 2024-04-20
finalize 2.15.0

YamatoSecurity:
@hitenkoku @fukusuket I decided to release 2.15.0 with the `windash` support to give people time to upgrade before we stop doing the `windash` conversion on the backend.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -6,6 +6,7 @@
 
 - Sigmaルールの `windash`フィールド修飾子 (例: `|contains|windash:`と`|contains|all|windash:`)に対応した。 (#1319) (@hitenkoku)
   - https://sigmahq.io/docs/basics/modifiers.html#windash
+  - 注意: 現在、バックエンドでは、以前のバージョンのHayabusaと互換性があるようにルール内の`windash`の使用を変換しているが、5月末ごろには、`windash`の使用をそのままにする予定なので、それまでにこのバージョンにアップデートしてください。
 
 **バグ修正:**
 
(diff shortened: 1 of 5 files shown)


hitenkoku (approved):
LGTM

YamatoSecurity:
@hitenkoku Sorry, could you check why this is failing? https://github.com/Yamato-Security/hayabusa/actions/runs/8761069797

fukusuket (commented):
This seems to be caused by the following changes :(
(It was probably compiled with `0.11.0` before, and now it's compiled with `0.10.x`)
- https://github.com/console-rs/dialoguer/blob/master/CHANGELOG.md#breaking
- https://github.com/console-rs/dialoguer/pull/277

In my environment, the issue was resolved by upgrading to version 0.11 or higher as shown below!
- https://github.com/Yamato-Security/hayabusa/blob/finalize-2.15.0/Cargo.toml#L22
- `dialoguer = "0.11.*"`

YamatoSecurity:
@fukusuket Thanks! It still wasn't working just by updating dialoguer in Hayabusa for me but when I updated it in hayabusa-evtx it now compiles without any error.

fukusuket (approved):
hayabusa evtx also uses dialoger, so this is probably the cause! I think it's better to update this as well:) LGTM🚀

https://github.com/Yamato-Security/hayabusa-evtx/blob/15f9f4a502d6b04fd6ee8c42dc3dc36a8b3e2bd1/Cargo.toml#L34

fukusuket (approved):
hayabusa evtx also uses dialoger, so this is probably the cause! I think it's better to update this as well:) LGTM🚀

https://github.com/Yamato-Security/hayabusa-evtx/blob/15f9f4a502d6b04fd6ee8c42dc3dc36a8b3e2bd1/Cargo.toml#L34

Issue #1342, opened 2024-05-07
Improving count rule's output


fukusuket:
https://github.com/Yamato-Security/hayabusa/pull/1341#issuecomment-2097170105

>1. Is it possible to put in the `Channel` and `EventID` info? When there are multiple values, we can separate them with ` ¦ `.
>2. `ExtraFieldInfo` is blank (`··`) but would nice to have it as `-` which we use for `n/a`. So `··` -> `· - ·`
>3. `EvtxFile` is also `-` but many users ask me how they can look up the evtx file from this alert so it would be nice to get this information as well. Again, separate by ` ¦ ` when there are multiple values.
>4. Is it possible to do the field key mapping? For example, change `SubStatus: 0xc000006a` to `SubStatus: WRONG PW`, etc...

fukusuket:
https://github.com/Yamato-Security/hayabusa/issues/1339#issuecomment-2092704365

>Or even better, if we could define in details: 'TgtUser: %TargetUserName% ¦ SrcIp: %IpAddress%' and get the following results: Count: 4 ¦ TgtUser: tanaka/Administrator/adsyncadmin/suzuki ¦ SrcIp: -

fukusuket:
https://github.com/Yamato-Security/hayabusa/pull/1341#issuecomment-2099492290

>- If top level Channel has multiple values, should it be an array or string?
>- If top level EventID has multiple values, should it be an array or string?
>- If each field under Details has multiple values, should it be an array or string?

YamatoSecurity:
@fukusuket I talked to someone who is more familiar with importing into elastic stack and he said that elastic will create multiple logs (documents) when arrays are used and it will make parsing more difficult so I think we should avoid arrays and keep everything as strings. We should probably keep the Channel and EventID information not in Details, but in the normal fields so that it makes searching for them easier.

fukusuket:
@YamatoSecurity 
Thank you for checking :) Is the expected JSON output format like the following?
- Top-level fields are strings joined with `|`. 
- Create fields such as `Count/IpAddress/SubStatus`... under the Details field and store values ​​in each.
```JSON
{
    "Timestamp": "2021-10-24 06:50:11.666 +09:00",
    "RuleTitle": "TEST_TITLE",
    "Level": "info",
    "Computer": "HOSTA ¦ HOSTB",
    "Channel": "Sysmon ¦ Security",
    "EventID": "4624 ¦ 4625",
    "RecordID": "",
[...]

PR #1357, opened 2024-06-02
feat: add support for Sigma Correlation Event count

### LINKED ISSUE #1337: Sigma correlations support: Event Count
@fukusuket Here is a (maybe) hard one for you. I'd like to start supporting sigma correlations starting with Event Count and Value Count types since we already have the logic for those with `| count`. We should probably start with `Event Count` as that is the easiest.

Both are explained here: https://github.com/SigmaHQ/sigma-specification/blob/version_2/Sigma_meta_rules.md

Event Count sample:
```
title: Many failed logins to the same computer
id: 0e95725d-7320-415d-80f7-004da920fc11
correlation:
  type: event_count
  rules:
    - e87bd730-df45-4ae9-85de-6c75369c5d29 # Logon Failure (Wrong Password)
    - 8afa97ce-a217-4f7c-aced-3e320a57756d # Logon Failure (User Does Not Exist)
  group-by:
    - Computer
  timespan: 5m
  condition:
    gte: 3
```

This is example uses multiple rules but at first we can just support one rule if that is easier.
Or you could create a generic rule with a new ID and the following logic to test:
```
logsource:
    product: windows
    service: security
detection:
    selection:
        Channel: Security
        EventID: 4625
```
This just detects any failed logon.

I think this should be the same as the following `count` rule:
```
logsource:
    product: windows
    service: security
detection:
    selection:
        Channel: Security
        EventID: 4625
    condition: selection | count() by Computer >= 3 
    timeframe: 5m
```

If there a 3 or more failed logons to the **same** Computer, then there will be an alert.

Things to consider:
* There *may* be multiple rules...  If it is difficult to implement checking multiple conditions from multiple rules then we might want to just start off by supporting one rule
* Rules may be defined by `name` instead of `ID` so we need a way to look this up internally. Info: https://github.com/SigmaHQ/sigma-specification/blob/version_2/Sigma_meta_rules.md#related-rules
* There may be multiple fields in `group-by`. I believe this means that, for example if `Computer` and `IpAddress` were defined in `group-by` then if the combination of the  **same** `ComputerName` **AND** the **same** `IpAddress` reaches 3 or more, then we should give an alert.

fukusuket:
## What Changed

- Closed #1337 

## Limitations of current implementation
The following points are not supported because they require more time to implement. 
I will consider addressing them as separate PR depending on priority.
- https://github.com/Yamato-Security/hayabusa/issues/1342
- [Multiple yaml in one file](https://github.com/SigmaHQ/sigma-specification/blob/version_2/Sigma_meta_rules.md#failed-logins-followed-by-successful-login)
- Multiple `Condition`(e.g.  100 > count > 1)
- Recursive Correlation rule

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -10,6 +10,7 @@
     - [Mimikatz Use](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/win_alert_mimikatz_keywords.yml)
 - デフォルトでは、適用可能なルールを持つ`.evtx`ファイルのみ読み込む。たとえば、さまざまなイベントログのディレクトリをスキャンしている場合でも、 `Channel: Security` を探すルールのみを有効にした場合、Hayabusaは`Security`以外のすべてのイベントログを無視します。ベンチマークでは、通常のスキャンで約10％、単一のルールでスキャンする場合は最大60％以上のパフォーマンス向上が得られる。チャネルに関係なくすべての`.evtx`ファイルを読み込みたい場合は、`csv-timeline` と `json-timeline` の `-a、--scan-all-evtx-files` オプションでこのフィルタリングをオフにすることができる。(#1318) (@fukusuket)
 - 注意: チャンネルフィルタリングは .evtx ファイルにのみ適用され、`-J, --json-input`オプションを使用してイベントログをJSONファイルから読み込む際に`-A`または`-a`を指定するとエラーが発生する。(#1345) (@fukusuket)
+- Sigma CorrelationのEvent Countに対応した。 (#1337) (@fukusuket)
 
 **改善:**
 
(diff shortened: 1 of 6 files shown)


fukusuket:
### Correlation

fukusuket:
### hayabusa-sample-evtx

fukusuket:
### hayabusa-sample-evtx
I confirmed that there are no differences in the result files as follows
#### csv-timeline
```
% ./hayabusa-old csv-timeline -d ../hayabusa-sample-evtx -w -D -u -n -q -C -o old.csv
% ./hayabusa-new csv-timeline -d ../hayabusa-sample-evtx -w -D -u -n -q -C -o new.csv
% diff new.csv old.csv
%
```

#### json-timeline
```
% ./hayabusa-old json-timeline -d ../hayabusa-sample-evtx -w -D -u -n -q -C -o old.json
% ./hayabusa-new json-timeline -d ../hayabusa-sample-evtx -w -D -u -n -q -C -o new.json
% diff new.json old.json
%
```
#### Integration-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/9342831069

hitenkoku (approved):
LGTM

YamatoSecurity (approved):

PR #1364, opened 2024-06-15
Embedded rules

hitenkoku:
## What Changed

- enabled embedded rule.

I would appreciate it if you could check it out when you have time

### DIFF
--- a/src/afterfact.rs
+++ b/src/afterfact.rs
@@ -2285,7 +2285,7 @@ mod tests {
                 us_time: false,
                 utc: false,
                 visualize_timeline: false,
-                rules: Path::new("./rules").to_path_buf(),
+                rules: Some(Path::new("./rules").to_path_buf()),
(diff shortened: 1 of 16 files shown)


fukusuket (approved):
I confirmed it works!! LGTM!!🚀
- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/powershell/powershell_script/posh_ps_tamper_windows_defender_set_mp. yml is still detected by Defender
- Binaries with ↑ included must not be detected by Defender (to include ↑, you have to disable Defender and compile it, w/o Defender)
- Without -r option and without rule/hayabusa and rule/sigma, scan by the rules in the binary.
- Without -r option, and with any rule under rule/hayabusa, scan for additional rules.
- Rules in the directory specified by the -r option are read from the file.

hitenkoku:
Close due to a policy of non-use.

PR #1366, opened 2024-06-15
updated cidr-utils crate to 0.6

hitenkoku:
## What Changed

- updated cidr-utils crate to 0.6

I would appreciate it if you could check it out when you have time


YamatoSecurity (approved):
@hitenkoku LGTM! Thanks so much!

PR #1405, opened 2024-08-28
doc: update --sort-events option description

fukusuket:
The -s option has been changed to sort-event in 2.17.0, so the README has been updated.


YamatoSecurity (approved):

PR #1411, opened 2024-09-03
Fix: Github Actions hayabusa-rule path

fukusuket:
## What Changed
The path to Actions in supported-modifier was incorrect and has been corrected.


YamatoSecurity (approved):
@fukusuket LGTM! Thanks!

PR #1417, opened 2024-09-19
feat(matchers): added fieldref modifier

hitenkoku:
## What Changed

- added `fieldref` modifier

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -1,5 +1,11 @@
 # 変更点
 
+## 2.18.0 [xxxx/xx/xx]
+
+**新機能:**
+
+- `fieldref`モディファイアに対応した。(#1409) (@hitenkoku)
+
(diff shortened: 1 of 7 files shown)


YamatoSecurity (approved):
@hitenkoku Thanks so much! I tested it and LGTM!

fukusuket (approved):
Sorry for the late confirmation.I confirmed following rule works!(Not to be detected with a value that does not exist) LGTM!!🚀
```
detection:
    selection_basic:
        Channel: Security
        EventID: 4624       # Logon Succesful
    selection_TokenImpersonationCharacteristics:
        LogonType: 9        # New Interactive
        ImpersonationLevel|contains: 1833 # It is actually %%1833 for Impersonation level of "Impersonation".
        SubjectUserName|fieldref: TargetUserName
    condition: selection_basic and selection_TokenImpersonationCharacteristics
```
There is also no difference between the main and the result.
https://github.com/Yamato-Security/hayabusa/actions/runs/10958330759

Issue #1281, opened 2024-02-16
timeline::metrics failing nestest tests


YamatoSecurity:
```
cargo install cargo-nextest
cargo nextest run
```
produces these results:
```
--- STDOUT:              hayabusa timeline::metrics::tests::test_evt_logon_stats ---

running 1 test
test timeline::metrics::tests::test_evt_logon_stats ... FAILED

failures:

failures:
    timeline::metrics::tests::test_evt_logon_stats

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 330 filtered out; finished in 0.04s


--- STDERR:              hayabusa timeline::metrics::tests::test_evt_logon_stats ---
thread 'timeline::metrics::tests::test_evt_logon_stats' panicked at src/detections/utils.rs:319:43:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

   Canceling due to test failure: 15 tests still running
        PASS [   0.078s] hayabusa timeline::computer_metrics::tests::test_computer_metrics_dsp_msg
        PASS [   0.079s] hayabusa timeline::timelines::tests::test_evt_logon_stats
        PASS [   0.082s] hayabusa timeline::timelines::tests::test_tm_logon_stats_dsp_msg
        PASS [   0.081s] hayabusa timeline::timelines::tests::test_tm_stats_dsp_msg
        PASS [   0.087s] hayabusa yaml::tests::test_all_exclude_rules_file
        PASS [   0.087s] hayabusa yaml::tests::test_all_noisy_rules_file
        PASS [   0.083s] hayabusa yaml::tests::test_critical_exact_level_read_yaml
        PASS [   0.082s] hayabusa yaml::tests::test_critical_level_read_yaml
        PASS [   0.033s] hayabusa yaml::tests::test_failed_read_yaml
        PASS [   0.078s] hayabusa yaml::tests::test_exclude_category_option_1opt
        PASS [   0.084s] hayabusa yaml::tests::test_default_level_read_yaml
        PASS [   0.148s] hayabusa options::profile::tests::test_load_profile
        PASS [   0.079s] hayabusa yaml::tests::test_exclude_category_option_multi_opt
        PASS [   0.068s] hayabusa yaml::tests::test_exclude_category_option_notfound
        PASS [   0.064s] hayabusa yaml::tests::test_exclude_deprecated_rules_file
[...]

PR #1418, opened 2024-09-25
metrics: handle None value unwrap panic

jmtaylor90:
## What Changed

- update metrics.rs to handle unwrap() panic on None values

## Evidence

This addresses #1281 which I ran into while running eid-metrics with the latest commit built. Though I don't know if this condition is being hit due to a different/larger issue.

### DIFF
--- a/src/timeline/metrics.rs
+++ b/src/timeline/metrics.rs
@@ -184,7 +184,7 @@ impl EventMetrics {
             let channel = if let Some(ch) =
                 utils::get_event_value("Channel", &record.record, &stored_static.eventkey_alias)
             {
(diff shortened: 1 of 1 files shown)


YamatoSecurity:
@jmtaylor90 Any reason why you closed this?

jmtaylor90:
> @jmtaylor90 Any reason why you closed this?

Hi, as I was double checking the work I realized that I was looking at a similar issue but different than the issue I was thinking my work was related to. I am going to see about fixing both and reopening.

YamatoSecurity:
@jmtaylor90 Thanks!

PR #1427, opened 2024-10-06
fix: add `logs` directory to the Windows package generated by release automation actions

fukusuket:
## What Changed
Sorry... I noticed there is no `logs` directory in the windows package generated by release automation actions ...
I fixed the GitHub Actions so that the `logs` directory is created in the windows package.

## Evidence
https://github.com/fukusuket/hayabusa/actions/runs/11197594221

スクリーンショット 2024-10-06 9 59 35

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,9 +93,9 @@ jobs:
         shell: pwsh
         run: |
           mkdir -p release-binaries
-          mkdir -p release-binaries/logs
           Copy-Item -Path ./target/release/hayabusa.exe -Destination release-binaries/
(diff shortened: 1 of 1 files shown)


YamatoSecurity:
@fukusuket I tested this on my Mac and it seems that the `logs` directory gets created when an error happens so I figure we could just leave it as it is without the directory being placed in the beginning and just create the directory when needed. (when an error occurs). What do you think?
Is there any part of the code that expects the directory to be there?

fukusuket:
@YamatoSecurity 
Thank you so much for checking! I see, Yes, it is better not to change it! I will close this!

PR #1500, opened 2024-11-17
feat: add file size next to file name under the progress bar

### LINKED ISSUE #1471: Add file size next to file name under the progress bar
Sometimes it takes quite a while to process a file because it is very big and we cannot give completely accurate progress bar progress as the .evtx file does not tell the amount of records. Some people may think that Hayabusa has hung so I want to make it easier to understand that it is taking time because the file size is big.

Before:
```
[00:00:29] 576 / 1,237 ⠤ [==================>                     ] 47%

"../hayabusa-samples/Microsoft-Windows-Sysmon%4Operational.evtx"
```

After:
```
[00:00:29] 576 / 1,237 ⠤ [==================>                     ] 47%

"../hayabusa-samples/Microsoft-Windows-Sysmon%4Operational.evtx (50.2 GB)"
```

We should use the same format for what we do when calculating `Total file size`. Ex: `17.3 KB`, `14.5 MB`, `12.2 GB`, etc...

fukusuket:
## What Changed

- Closed #1471 

### Integration-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/11874529743


fukusuket:
スクリーンショット 2024-11-17 9 45 12

YamatoSecurity (approved):
@fukusuket LGTM! Thanks so much for such a super quick PR!

PR #1514, opened 2024-11-28
Prepare for rust 2024

YamatoSecurity:
As per https://blog.rust-lang.org/2024/11/27/Rust-2024-public-testing.html , I tested to see if Hayabusa was compatible with the new Rust edition 2024 that will be release next Feb. Good news is that all works without any warnings or errors.
Since the changes that it made are backwards compatible to the current 2021 edition I figured we might as well adapt to the new coding style.

@fukusuket Whenever you get time, could you check over these?


YamatoSecurity:
@fukusuket Ahh, it seems to only work with the cargo nightly build i think. I will close this for now and we can think about migrating to 2024 edition next year.

PR #1517, opened 2024-11-29
Update to 3.0.0 dev build

YamatoSecurity:
I think the clippy warnings have become more strict so we were getting compile errors.


YamatoSecurity:
@fukusuket Do you know why the code coverage is failing?

YamatoSecurity:
@fukusuket Looks like this error: https://github.com/rust-lang/rust/issues/133606
I suppose we can just wait a few days until they fix it...

fukusuket:
@YamatoSecurity 
I see..., If the `nightly` build is not required, would the Actions succeed if we change the following to `stable`?
https://github.com/Yamato-Security/hayabusa/blob/95c86d771bb1e65b0e86c93d2ab7964406dd56df/.github/workflows/coverage.yml#L17-L18

fukusuket (approved):
LGTM!!🚀

PR #1552, opened 2025-01-23
feat: add `-X, --remove-duplicate-detections` option to `eid-metrics` and `logon-summary`

### LINKED ISSUE #1478: Invalid: `-x` option in `logon-summary` and `eid-metrics` will result in duplicate results
Turning on `-x` to carve records from slack will often result in duplicate events being read, however, `logon-summary` and `eid-metrics` does not do any filtering for duplicate events. I think we need to keep track of timestamp, etc.. information for events only when `-x` is being used and filter on duplicate events before giving results in order to give more accurate results.

fukusuket:
## What Changed
- Closed #1478 

### Integration-Test
- https://github.com/Yamato-Security/hayabusa/actions/runs/12927978824

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -2,6 +2,10 @@
 
 ## 3.1.0 [2025/02/22] - Ninja Day Release
 
+**新機能:**
+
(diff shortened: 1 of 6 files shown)


fukusuket:
### eid-metrics

fukusuket:
### logon-summary

YamatoSecurity (approved):

PR #1560, opened 2025-01-30
feat: add PowerShellCore/Operational to extract-base64

### LINKED ISSUE #1558: Support PowerShell Core Logs in extract-base64 command
I believe that the extract-base64 command only looks at Windows Powershell but as the usage of PowerShell Core is going to increase in the future we should also support `PowerShellCore/Operational` as well.

Ref: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_logging_windows?view=powershell-7.5

fukusuket:
## What Changed

-  Closed #1558 

### Integration-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/13052625886


YamatoSecurity (approved):

PR #1579, opened 2025-02-14
Add RuleID by default

YamatoSecurity:
I added `RuleID` by default to all profiles except minimal as it is really useful for looking up sigma rules.

I also updated the search criteria part of the Japanese readme for timeline explorer. Could you check it?


fukusuket (approved):
LGTM!!🚀

PR #1588, opened 2025-02-16
refactor: use Default trait to reduce unnecessary initialization codes in StoredStatic

fukusuket:
## What Changed
- refactor: use Default trait to reduce unnecessary initialization codes
   - Modifications that change storedstatic, such as adding commands or options, require that other unrelated files be updated as well. (Because there are many places where storedstatic is used in the test code.)
   - By using the Default Trait, we can reduce the number of unrelated files that need to be updated in the future when adding commands and options.
### Integration-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/13352480679

### Release automation
https://github.com/Yamato-Security/hayabusa/actions/runs/13352514055


YamatoSecurity (approved):
@fukusuket I ran some tests with `csv-timeline` and get identical results so LGTM!

Issue #1606, opened 2025-02-28
[bug] Counting `expand` rules as valid rules even when `config/expand/xxx.txt` does not exists


fukusuket:
**Describe the bug**
Counting `expand` rules as valid rules even when `config/expand/xxx.txt` does not exists.

**Step to Reproduce**
```yml
status: test
description: test
author: TEST
date: 2024-12-07
logsource:
    product: windows
    service: security
    definition: TEST
detection:
    selection:
        Channel: Security
        EventID: 4624
        LogonType|expand: '%LogonType%'
    condition: selection
level: informational
```
`./hayabusa csv-timeline -d ../hayabusa-sample-evtx -r test.yml -w -q`

**Expected behavior**
```
Test rules: 1 (100.00%)

Expand rules: 1 (100.00%)
Enabled expand rules: 0 (0.00%)

Total detection rules: 0

Creating the channel filter. Please wait.

Evtx files loaded after channel filter: 248
Detection rules enabled after channel filter: 0
```

**Actual behavior**
```
Test rules: 1 (100.00%)

Expand rules: 1 (100.00%)
Other rules: 1
Total detection rules: 1

Creating the channel filter. Please wait.

Evtx files loaded after channel filter: 248
Detection rules enabled after channel filter: 1
```

**Environment**

- OS: macOS Sequoia 15.3.1
- hayabusa version 3.1.0

**Additional context**
N/A

PR #1608, opened 2025-03-01
test: add `live-response` package result count check in Integration-Test

### LINKED ISSUE #1597: Automation for checking if the live response version gives the same results
It would be nice to automate checks to see if the normal version of Hayabusa and the live response version load the same rules and produce the same results.

fukusuket:
## What Changed
- Closed #1597

## Test Specification
On Windows, there is a difference in the number of results due to Windows Defender. 
Therefore, compare the number of  `wc -l` only on Ubuntu and macOS.

### Integraion-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/13604064714
(Tests will fail until https://github.com/Yamato-Security/hayabusa/issues/1606  is fixed)

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -200,3 +200,50 @@ jobs:
 
       - name: set-default-profile
         run: cd main && ./hayabusa set-default-profile -p verbose -q
+
+  integration-test-live-response-results:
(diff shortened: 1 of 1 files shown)


YamatoSecurity (approved):
@fukusuket LGTM! Thanks!!

PR #1612, opened 2025-03-01
fix: extract-base64 double encoding check

### LINKED ISSUE #1607: Double encoding not properly detected with `extract-base64` command
@fukusuket This is low priority but when you have time could you check the `extract-base64` output with the data from the contest last weekend. I think there was double encoding in the results but it was listed as `N`.

fukusuket:
## What Changed
- Closed #1607 

### Integraion-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/13605924827


YamatoSecurity (approved):

PR #1637, opened 2025-03-15
fix: improve logging for invalid rule

### LINKED ISSUE #1601: [bug] Bad rule is not logged to error log
**Describe the bug**
I was testing to see what fields in rules are required and in what scenarios a rule will not be processed by hayabusa. I came across a few scenarios where hayabusa will filter out the rule (b/c its bad) but it will not log an error while other cases it will. 

Scenarios where rule is filtered out but no error is recorded **(bug)**
1) If level is missing from the rule
2) If level has a bad value
3) if status has a bad value

A few scenarios where rule is filtered out but errors are recorded **(works as expected)**
1) detection is missing from the rule
2) condition is missing from detection section of rule
3) invalid command after pipe in detection section. (ex. ImagePath|badvalue: 'test')

**Expected behavior**
Rules that cannot be processed for whatever reason should be logged to the error log. The error should include the path to the rule and the reason why just like they currently do. This should work for live response as well

fukusuket:
## What Changed

- Closed #1601 

### Integrarion-Test
https://github.com/Yamato-Security/hayabusa/actions/runs/13871367228

### DIFF
--- a/src/yaml.rs
+++ b/src/yaml.rs
@@ -224,7 +224,7 @@ impl ParseYaml {
                 }
             }
         } else {
-            let mut entries = fs::read_dir(path)?;
+            let mut entries = fs::read_dir(&path)?;
             yaml_docs = entries.try_fold(vec![], |mut ret, entry| {
                 let entry = entry?;
                 // フォルダは再帰的に呼び出す。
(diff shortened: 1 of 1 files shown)


fukusuket:
### Test
```
% cat ./rules/hayabusa/invalid.yml
author: TEST
date: 2022/03/22

title: Invalid Rule
description:
id: f03e34c4-6432-4a30-9ae2-76ae6329399a
status: invalid
logsource:
    product: windows
    service: sysmon
detection:
    selection_basic:
        Channel: Microsoft-Windows-Sysmon/Operational
        EventID: 2
    condition: selection_basic
falsepositives:
tags:
references:
ruletype: Hayabusa
```

YamatoSecurity:
@fukusuket Thanks!
The rule above has a missing level and `status: invalid` but the error message says `Failed to parse rule:/Users/fukusuke/Hayabusa/hayabusa-3.1.1-mac-aarch64/rules/hayabusa. Invalid level:N/A`

How hard would it be to print all the wrong things with the rule?
Ex: `Missing: level ¦ Invalid: status ¦ Missing: xxxx ¦ Invalid: xxxx ¦ Invalid: xxx`

fukusuket:
@YamatoSecurity 
Thank you for checking! Sounds good! I'll look into :)

fukusuket:
@YamatoSecurity 
I have a question! 
In the following document, is it correct that the fields marked **Required** are the target of this new check?
- https://github.com/Yamato-Security/hayabusa-rules/tree/main?tab=readme-ov-file#author-section

YamatoSecurity:
@fukusuket Good question! The `hayabusa-rules` documentation is a little old so lets make these fields required:
```
author
date
title
id
level
status
detection
```

YamatoSecurity:
@fukusuket I realized the hayabusa-rules readme was correct! 笑
Lets also add `ruletype` to the required field.

fukusuket:
I am closing this Pull Request to open a new one for implementing the updated check specification!💪

PR #1665, opened 2025-06-12
fix: support access System32 dir without redirect, so we can run 32 bit exe on 64 bit os

akkuman:
support access System32 dir without redirect, so we can run 32 bit exe on 64 bit os

ref:
- https://learn.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-wow64disablewow64fsredirection
- https://github.com/Yamato-Security/hayabusa/issues/478
- https://github.com/Yamato-Security/hayabusa/pull/482

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,6 +926,7 @@ dependencies = [
  "uuid",
  "walkdir",
  "wildmatch",
+ "winapi",
  "yaml-rust2",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ rust-embed={version = "8.7.2", features = ["include-exclude", "debug-embed"]}
 encoding_rs = "0.8.35"
 walkdir = "2.5.0"
 uuid = { version = "1.11.0", features = ["v4"] }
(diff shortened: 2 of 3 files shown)


YamatoSecurity:
@akkuman Thanks for the PR! It is failing our build checks, so could you look at this?
Might be able to fix with `cargo fmt -- --check`

akkuman:
Ok, I'll test the suggested improvements given by copilot in my free time and see if they work as expected. Once it's done, I'll commit and push

akkuman:
@YamatoSecurity I tried the implementation given by copilot, 
```rust
        let _guard = if !self.is_matched_architecture_and_binary() {
            Wow64RedirectionGuard::new()
        } else {
            None
        };
```
 will report an error, which is necessary to implement RAII. If I change it to
```rust
let _guard = if !self.is_matched_architecture_and_binary() {
    Wow64RedirectionGuard::new()
} else {
    Ok(Wow64RedirectionGuard {
        #[cfg(target_os = "windows")]
        old_state: std::ptr::null_mut(),
    })
};
```
, it will cause old_state to be a null pointer. Calling Wow64RevertWow64FsRedirection will also report an error.

PR #1695, opened 2025-09-17
fix: JSON timeline flag in geo IP lookup

### LINKED ISSUE #1693: [bug] GeoIP lookup not working with JSON output
@fukusuket When you have time, could you take a look at this?
In the past, I remember the `-G` geo IP location lookup would work with JSON output as well but for some reason on 3.4.0+ it is not working. (I only tested with main and 3.4.0 so it could be broken in prior versions as well)
When I create a CSV file with `-G /opt/homebrew/var/GeoIP/`, it adds the `SrcASN`, `SrcCountry`, etc... fields and populates them, although with the sample evtx files there are not a lot of results besides `Local` and `Private`. I tested that it does seem to be working with global IP addresses in other scenarios.

However, when I save to JSON/L, I do not see any `SrcASN`, `SrcCountry`, etc.. fields at all, nor the results of the ASN, city, country that should be there.

fukusuket:
## What Changed

- Closed #1693 

### Integration-Test
- https://github.com/Yamato-Security/hayabusa/actions/runs/17794152661

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -9,6 +9,7 @@
 **バグ修正:**
 
 - `-T, --visualize-timeline`は、`-s, --sort`が指定されていない場合、正しくない結果を出力するため、`-T`を使用する際には`-s`の指定を必須とした。 (#1690) (@yamatosecurity)
+- GeoIP検索が`json-timeline`で機能していなかった。 (#1693) (@fukusuket)
 
 ## 3.5.0 [2025/08/16] - Obon Release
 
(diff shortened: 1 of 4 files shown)


fukusuket:
```
% ./hayabusa json-timeline -d ../data/windows/all-evtx/server -G ../data/maxmind -o timeline-new.json -w -C -p super-verbose
% cat timeline-new.json | grep TgtASN | sort | uniq
        "TgtASN": "Akamai International B.V.",
        "TgtASN": "AKAMAI-AS",
        "TgtASN": "AMAZON-02",
        "TgtASN": "AMAZON-AES",
        "TgtASN": "ANEXIA Internetdienstleistungs GmbH",
        "TgtASN": "ASN-APPNEX",
        "TgtASN": "Cloudflare London, LLC",
...
```

YamatoSecurity (approved):

PR #1711, opened 2025-10-28
feat: add checksum validation option 

### LINKED ISSUE #1709: Add checksum validation option
Under `General Options`, I would like to add an option to validate checksums for users who want to ignore possibly corrupted events.

`-V, --validate-checksums  Enable checksum validation`

This should be added to at least `csv-timeline` and `json-timeline`. If it is not too difficult, we can also add to the other commands that evtx files as well.

@fukusuket Could I ask you to do this one?

fukusuket:
## What Changed

- Closed #1709 

### Integration-Test 
- https://github.com/Yamato-Security/hayabusa/actions/runs/18876529102

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -2,6 +2,10 @@
 
 ## 3.7.0 [xxxx/xx/xx]
 
+**新機能:**
+
+- `csv-timeline`および`json-timeline`コマンドでチャンクヘッダーのチェックサムを確認する`-V, --validate-checksums`オプションを追加した。 (#1709) (@fukusuket)
+
 **バグ修正:**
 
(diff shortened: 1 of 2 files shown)


YamatoSecurity:
@fukusuket Thanks! I think we fixed the infinite loop bug with this PR: https://github.com/Yamato-Security/hayabusa/pull/1710

I tested out the `-V` usage but it doesn't seem that Hayabusa is ignoring events that have invalid checksums.
For example, when I run normal against some data I get 1,112,173 events.
When I add the `-x` option to recover records, it increases to 1,112,329, however, many of these records are broken and I am pretty sure the checksums won't be accurate.

When I run `-V -x` together, I expected Hayabusa to ignore all or most of these recovered records but it returns with the same 1,112,329 events.

I also tried running on various other sample data with and without `-V` expecting a difference in events with hits but they always come back the same. 

Could you double check that Hayabusa will ignore any events that have an invalid checksum when `-V` is enabled?

PR #1747, opened 2026-04-28
fix: resolve clippy warnings to unblock CI

Fuzzdkk:
## Summary

`cargo clippy --all-targets -- -D warnings` has been failing on `main` since the clippy 1.95 bump. The Rust workflow has been red on every PR for around two weeks, including dependabot ones. This fixes the 7 lints so CI goes back to green.

Changes:

- `src/afterfact.rs`: 3x `sort_by` -> `sort_by_key`
- `src/detections/rule/count.rs`: `sort_by` -> `sort_by_key`
- `src/filter.rs`: collapse `Yaml::String` + key check into a match guard
- `src/timeline/search.rs`: `sort_unstable_by` -> `sort_unstable_by_key`
- `src/yaml.rs`: collapse the `date` arm regex check into a match guard

All seven changes are equivalent to the originals. The sort keys (`i128`, `DateTime<Utc>`, `usize`) are all `Copy + Ord`. The two collapsed matches fall through to the existing `_ => {}` arm when the guard fails, matching the original do-nothing branch.

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `rustfmt --check` clean on all 5 files
- [x] `cargo test --lib` for every touched module (afterfact, detections::rule::count, filter, yaml) passes

### DIFF
--- a/src/afterfact.rs
+++ b/src/afterfact.rs
@@ -1488,7 +1488,7 @@ fn _print_detection_summary_by_computer(
             .filter(|a| a.0.as_str() != "-")
             .collect();
 
-        sorted_detections.sort_by(|a, b| (-a.1).cmp(&(-b.1)));
+        sorted_detections.sort_by_key(|a| -a.1);
 
         // html出力は各種すべてのコンピュータ名を表示するようにする
         if stored_static.html_report_flag {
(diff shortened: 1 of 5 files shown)


YamatoSecurity (approved):
@Fuzzdkk LGTM! Thanks for fixing the clippy errors!

PR #1760, opened 2026-06-19
fix: progress bar broken on large scans (literal `\r\n` in indicatif template)

### LINKED ISSUE #1759: Progress bar redraw breaks (prints on new lines) during large scans — literal `\r\n` in the indicatif template
## Summary
On large scans the scan progress bar stops redrawing in place and instead prints every tick on a **new line**, with a blank line between each, flooding the terminal:

```
[00:20:20] 18,713 / 47,925 ⠁ [===============>            ] 39%

[00:20:22] 18,738 / 47,925 ⠂ [===============>            ] 39%

[00:20:23] 18,809 / 47,925 ⠲ [===============>            ] 39%
...
```

Small scans render the bar correctly (single in-place line); the corruption only appears on large scans.

## Root cause
The progress template in `src/main.rs` (both the `no_color` and colored variants, lines ~2022 / ~2027) hard-codes literal carriage returns:

```rust
"[{elapsed_precise}] {human_pos} / {human_len} {spinner} [{bar:40}] {percent}%\r\n\r\n{msg}"
```

`indicatif` owns cursor/carriage-return handling and splits templates into lines **only on `\n`**, so the `\r`s become line *content* (a stray trailing `\r` on the bar line and a lone-`\r` "blank" line). On every tick the bar redraws by moving the cursor up N lines, which requires knowing how many **terminal rows** each line occupies (`ceil(visible_width / term_width)`). The stray `\r`s corrupt that width/row accounting, and once a line **wraps** indicatif moves up the wrong number of rows and re-prints the whole block below the old one instead of overwriting it.

## Why only on large scans
Both wrapping triggers scale with scan size:
- **Counters grow:** `18,713 / 47,925` is far wider than `18 / 50`; the colored line (`[elapsed] huge/huge {spinner} [40-char bar] NN%`) is ~80 cols and big numbers push it past the terminal width → it wraps.
- **`{msg}` is the file path** (`pb.set_message(...)`); large scans hit longer/deeper paths that also wrap.

Short scans keep every line within one terminal row, so the redraw stays correct.

## Fix
Use `\n` instead of literal `\r\n` in both templates (indicatif-idiomatic) — e.g. `…{percent}%\n\n{msg}`. The visible layout (bar, blank line, message) is unchanged; indicatif then manages line endings and wrapping itself.

Affected: `src/main.rs` (progress-bar setup, ~lines 2021–2040). PR to follow.

Shirofune-Security:
## Summary
Fixes the scan progress bar printing every tick on a **new line** (with a blank line between) during **large scans**, instead of redrawing in place.

```diff
-"...{percent}%\r\n\r\n{msg}"
+"...{percent}%\n\n{msg}"
```

## Cause
The progress template in `src/main.rs` (both `no_color` and colored variants, ~lines 2022/2027) embeds **literal carriage returns** (`\r\n\r\n`). `indicatif` owns cursor/CR handling and splits templates into lines **only on `\n`**, so the `\r`s become line *content* (a stray trailing `\r` on the bar line and a lone-`\r` "blank" line). That corrupts indicatif's per-line width/row accounting; once a line **wraps** — which large scans trigger because the counters (`18,713 / 47,925`) and the `{msg}` file path grow past the terminal width — indicatif moves the cursor up by the wrong number of rows and re-prints the block below the old one instead of overwriting it. Small scans keep each line within one row, so the redraw stays correct there.

Example of the broken output:
```
[00:20:20] 18,713 / 47,925 ⠁ [===============>            ] 39%

[00:20:22] 18,738 / 47,925 ⠂ [===============>            ] 39%
...
```

## Fix
Use `\n\n` (indicatif-idiomatic) in both templates. The visible layout — bar, blank line, message — is unchanged; indicatif now manages line endings and wrapping itself. (rustfmt then wraps the trailing `.to_string()` onto its own line.)

## Verification
With Rust 1.95.0: `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt -- --check`, and `cargo test` (**437 passed / 0 failed**) all pass. Template-string-only change, no logic.

Closes #1759.


YamatoSecurity (approved):
LGTM

PR #1765, opened 2026-06-19
test: serialize tests that race on global HTML_REPORTER / PIVOT_KEYWORD state

### LINKED ISSUE #1764: Flaky lib tests: test_output_profile / test_add_md_data / insert_pivot_keyword_* race on global lazy_static state under parallel cargo test
## Summary
Several **library** unit tests intermittently fail under the parallel test harness because they share **process-global `lazy_static` state** without serialization. Two independent groups:

- **`HTML_REPORTER`** (`src/options/htmlreport.rs`): `detections::utils::tests::test_output_profile` and `options::htmlreport::tests::test_add_md_data` each `clear()` → populate → read the global `HTML_REPORTER.md_datas`. When they run on different threads, one test's entry/`clear()` interleaves with the other's assertion.
- **`PIVOT_KEYWORD`** (`src/options/pivot.rs`): the 7 `insert_pivot_keyword_*` tests each `clear()` → load → insert → assert the global `PIVOT_KEYWORD`, racing with each other.

## Reproduction
```
# the two HTML_REPORTER tests, forced onto 2 threads:
for i in $(seq 1 40); do cargo test --lib -- test_output_profile test_add_md_data --test-threads=2; done
# => ~35/40 runs FAIL: test_output_profile panics at utils.rs:1134
#    `assertion failed: expect.keys().any(|x| x == k)`  (a foreign "AddTest {#add_test}" key from test_add_md_data)

# full lib suite, default parallelism:
for i in $(seq 1 12); do cargo test --lib; done
# => ~6/12 runs FAIL (a mix of the HTML_REPORTER and insert_pivot_keyword_* tests)
```
Both groups pass reliably with `--test-threads=1`.

## Root cause
The globals (`HTML_REPORTER: RwLock<HtmlReporter>`, `PIVOT_KEYWORD: RwLock<IndexMap<..>>`) are process-wide. Tests that mutate them are not mutually exclusive, so under the default multi-threaded harness one test observes another's writes/clears.

## Proposed fix
Add a per-global `#[cfg(test)] static …_TEST_LOCK: Mutex<()>` and have each test that touches the global hold it for the whole test body (`.lock().unwrap_or_else(|e| e.into_inner())` to tolerate poisoning). Zero new dependencies; serializes only the affected tests, leaving the rest parallel.

## Note (separate, not covered here)
The **binary** tests `tests::test_same_file_output_*_exit` (`src/main.rs`) are a *different* flaky family — they race on the global `STORED_STATIC`/`STORED_EKEY_ALIAS` and shared fixed output filenames. Out of scope for this issue; mentioned for awareness.

PR to follow.

Shirofune-Security:
## Summary
Fixes intermittent failures of several **library** unit tests that share process-global `lazy_static` state and run concurrently under the default test harness.

Two independent groups:
- **`HTML_REPORTER`** — `detections::utils::tests::test_output_profile` and `options::htmlreport::tests::test_add_md_data` each `clear()` → populate → read the global `HTML_REPORTER.md_datas`. Concurrently, one test's entry/`clear()` interleaves with the other's assertion (`test_output_profile` panics at `utils.rs:1134`, seeing a foreign `"AddTest {#add_test}"` key).
- **`PIVOT_KEYWORD`** — the 7 `options::pivot::tests::insert_pivot_keyword_*` tests each `clear()` → load → insert → assert the global, racing with one another.

## Fix
Add a per-global, test-only serialization lock and have each affected test hold it for its whole body:
```rust
#[cfg(test)]
pub(crate) static HTML_REPORTER_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
// ...
let _lock = HTML_REPORTER_TEST_LOCK.lock().unwrap_or_else(|e| e.into_inner());
```
(`unwrap_or_else(into_inner)` tolerates poisoning so a panic in one test doesn't cascade.) **Zero new dependencies**; only the tests that touch each global are serialized — everything else stays parallel.

## Verification (Rust 1.95.0, against the real deps)
- HTML_REPORTER tests forced onto 2 threads, 50 runs: **0 failures** (pre-fix: **35/40** failed).
- Full `cargo test --lib`, default parallelism, 20 runs: **0 failures** (pre-fix: **6/12** failed); **420 passed**.
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check`: clean.

This is a test-only change (no runtime behavior change), so no CHANGELOG entry.

## Out of scope (separate pre-existing issue)
The **binary** tests `tests::test_same_file_output_*_exit` (`src/main.rs`) are a *different* flaky family — they race on the global `STORED_STATIC`/`STORED_EKEY_ALIAS` and shared fixed output filenames. Not touched here; noted in #1764 for awareness.

Closes #1764.

### DIFF
--- a/src/detections/utils.rs
+++ b/src/detections/utils.rs
@@ -1107,6 +1107,10 @@ mod tests {
 
     #[test]
     fn test_output_profile() {
+        // Serialize against other tests that mutate the global HTML_REPORTER.
+        let _html_reporter_lock = crate::options::htmlreport::HTML_REPORTER_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
         HTML_REPORTER.write().unwrap().md_datas.clear();
         let stored_static = StoredStatic::create_static_data(Some(Config {
             action: Some(Action::CsvTimeline(CsvOutputOption {
--- a/src/options/htmlreport.rs
+++ b/src/options/htmlreport.rs
@@ -23,6 +23,12 @@ lazy_static! {
     pub static ref HTML_REPORTER: RwLock<HtmlReporter> = RwLock::new(HtmlReporter::new());
 }
 
+/// Serializes the tests that mutate the process-global `HTML_REPORTER`
+/// (clear/populate/read) so they don't race under the parallel test harness.
+/// Each such test must hold this lock for its whole body.
+#[cfg(test)]
+pub(crate) static HTML_REPORTER_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
 #[derive(Clone)]
 pub struct HtmlReporter {
     pub section_order: Nested<String>,
@@ -202,7 +208,7 @@ mod tests {
 
     use nested::Nested;
 
-    use super::{HTML_REPORTER, img_to_base64};
+    use super::{HTML_REPORTER, HTML_REPORTER_TEST_LOCK, img_to_base64};
     use crate::{
         detections::configs::{
             Action, Config, CsvOutputOption, JSONOutputOption, OutputOption, StoredStatic,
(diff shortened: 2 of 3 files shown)


YamatoSecurity (approved):
LGTM

PR #1767, opened 2026-06-19
test: serialize bin tests that share output filenames (overwrite-*)

### LINKED ISSUE #1766: Flaky bin tests: test_same_file_output_*_exit race on shared output filenames under parallel cargo test
## Summary
Several **binary** unit tests in `src/main.rs` intermittently fail under the parallel test harness because an `*_exit` test and its non-`_exit` sibling **write to the same fixed output filename with opposite expectations**.

Example pair (both use `overwrite-metric.csv`):
- `test_same_file_output_metric_csv_exit` — no `clobber` → hayabusa refuses to overwrite the existing file → asserts `meta.len() == 0`.
- `test_same_file_output_metric_csv` — `clobber: true` → hayabusa overwrites → asserts `meta.len() != 0`.

Run concurrently, the clobber sibling writes ~92 bytes to the file the `*_exit` test expects to stay empty, so the `*_exit` assertion fails:
```
thread 'tests::test_same_file_output_metric_csv_exit' panicked at src/main.rs:3484:
assertion `left == right` failed
  left: 92
 right: 0
```

## Affected tests (10, sharing 5 filenames)
`test_same_file_output_csv_exit`/`test_overwrite_csv`/`test_overwrite_json` (`overwrite.csv`/`overwrite.json`), `test_same_file_output_json_exit`, `test_same_file_output_metric_csv{,_exit}` (`overwrite-metric.csv`), `test_same_file_output_logon_summary_csv{,_exit}` (`overwrite-metric-successful.csv`), `test_same_file_output_computer_metrics{_csv,_exit}` (`overwrite-computer-metrics.csv`).

## Root cause
Pure filesystem collision on the shared filename. The overwrite check uses the `stored_static` **argument** and returns *before* any analysis (so the process-global `STORED_STATIC` and detection are not involved for the `*_exit` path).

## Reproduction
```
for i in $(seq 1 6); do cargo test --bin hayabusa; done   # ~6/6 fail
```
Passes with `--test-threads=1`.

## Proposed fix
Serialize the 10 output-file tests with a shared `#[cfg(test)] static FILE_OUTPUT_TEST_LOCK: Mutex<()>` held for each test body (`unwrap_or_else(into_inner)` to tolerate poisoning). Zero new dependencies. (Same approach as the lib-test fix in #1765.)

This is the bin-test family noted as out-of-scope in #1764. PR to follow.

Shirofune-Security:
## Summary
Fixes intermittent failures of the **binary** output-file tests in `src/main.rs`. An `*_exit` test and its sibling write the **same fixed output filename** with opposite expectations:

- `test_same_file_output_metric_csv_exit` — no `clobber` → refuses overwrite → asserts `meta.len() == 0`.
- `test_same_file_output_metric_csv` — `clobber: true` → overwrites → asserts `meta.len() != 0`.

Run concurrently, the clobbering sibling writes ~92 bytes to `overwrite-metric.csv` while the `*_exit` test asserts it is empty → `left: 92, right: 0` (`main.rs:3484`). Five such filename pairs across 10 tests.

## Root cause
A pure filesystem collision. The no-clobber overwrite check reads the `stored_static` **argument** and `return`s before any analysis — so the process-global `STORED_STATIC` / detection are not involved; only the shared filename is.

## Fix
Serialize the 10 output-file tests with a shared, test-only lock held for each test body:
```rust
static FILE_OUTPUT_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
// ...at the start of each affected test:
let _file_output_lock = FILE_OUTPUT_TEST_LOCK.lock().unwrap_or_else(|e| e.into_inner());
```
**Zero new dependencies**; only these 10 tests are serialized, everything else stays parallel. Same approach as the lib-test fix in #1765.

## Verification (Rust 1.95.0)
`cargo test --bin hayabusa`, default parallelism: **6/6 runs failed pre-fix → 0/15 failed post-fix** (17 passed). `cargo clippy --all-targets -- -D warnings` + `cargo fmt --check`: clean.

Test-only change. Completes the flaky-test cleanup started in #1765 (this was the bin family noted as out-of-scope in #1764).

Closes #1766.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3045,6 +3045,13 @@ mod tests {
         timeline::timelines::Timeline,
     };
 
+    /// Serializes the tests that exercise output-file overwrite behavior. They
+    /// share fixed output filenames (e.g. `overwrite-metric.csv`) between the
+    /// `*_exit` (no-clobber, expects the file untouched) and clobber variants,
+    /// so running them concurrently lets one test write the file the other
+    /// asserts is empty. Each such test holds this lock for its whole body.
+    static FILE_OUTPUT_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     fn create_dummy_stored_static() -> StoredStatic {
         StoredStatic::create_static_data(Some(Config {
             action: Some(Action::CsvTimeline(CsvOutputOption {
@@ -3164,6 +3171,10 @@ mod tests {
 
     #[test]
     fn test_same_file_output_csv_exit() {
+        // Serialize against the sibling test that shares this output filename.
+        let _file_output_lock = FILE_OUTPUT_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
         // Create an empty file first.
         let mut app = App::new(None);
         File::create("overwrite.csv").ok();
@@ -3204,6 +3215,10 @@ mod tests {
 
     #[test]
     fn test_overwrite_csv() {
+        // Serialize against the sibling test that shares this output filename.
+        let _file_output_lock = FILE_OUTPUT_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
         // Create an empty file first.
         let mut app = App::new(None);
         File::create("overwrite.csv").ok();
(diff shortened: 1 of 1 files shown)


Shirofune-Security:
Thanks @Copilot — good catch, fixed in the latest commit. `test_overwrite_json` now does `File::create("overwrite.json")` (its actual output path) instead of `overwrite.csv`, so it properly exercises the JSON overwrite-with-`clobber` path, no longer touches the CSV tests' shared file, and no longer leaks `overwrite.csv` (the cleanup already only removed `overwrite.json`).

Verified: `cargo test --bin hayabusa` stays green (0/8 runs failed, 17 passed) and `overwrite.csv` is no longer produced.

YamatoSecurity (approved):
LGTM

PR #1782, opened 2026-06-27
chore: bump evtx to 0.9.8 and update crates

Shirofune-Security:
## Summary
- Bumps **hayabusa-evtx `0.9.7` → `0.9.8`** (rev `c1367df`, the merge of [hayabusa-evtx#90](https://github.com/Yamato-Security/hayabusa-evtx/pull/90)).
- Runs `cargo update` on the remaining dependencies (anyhow, bstr, cc, chacha20, js-sys, shlex, uuid, wasm-bindgen, …).

## Why the evtx bump matters
hayabusa-evtx#90 fixes JSON serialization of a **present-but-empty** named `<Data Name="X"></Data>` element: it now renders as `""` (present, empty) instead of `null` (which this serializer uses for an *absent* field).

This matters for Sigma `field: null` matching, which should match only an **absent** field. Before the fix, a present-but-empty field (e.g. `ApplicationPath` on a Windows Firewall EventID-2004 *service* rule) serialized to `null` and was wrongly treated as absent, so a `filter_main_null: ApplicationPath: null` filter matched and **suppressed a legitimate detection**. With `0.9.8` the field is `""`, the filter no longer matches, and the rule fires correctly — matching the behavior of the 0.12.x evtx line.

## Verification
- `cargo build` clean (no warnings) against evtx `0.9.8` + the updated crates.
- `cargo fmt --check` clean.
- Smoke scan parses a sample evtx and produces detections normally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "arraydeque"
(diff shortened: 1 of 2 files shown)


YamatoSecurity (approved):
LGTM

PR #1786, opened 2026-06-28
docs: fix paragraph-introduced lists not rendering on the docs site

Shirofune-Security:
## Summary

Several lists in the docs site followed an intro sentence with **no blank line** between them, e.g. on the Downloads page:

```
We provide binaries for the following architectures:
- Linux ARM 64-bit GNU (...)
- Linux Intel 64-bit GNU (...)
```

GitHub-Flavored Markdown renders that as a list, but the docs site uses **MkDocs / Python-Markdown**, which requires a blank line before a list and otherwise collapses the whole thing into one run-on paragraph.

## Fix

Insert a blank line before every **top-level list item that directly follows a top-level paragraph**, across all 15 languages (605 insertions). The change is intentionally conservative — content inside code fences, YAML front matter, headings, tables, and nested/indented lists is left untouched — so it only adds the blank lines Python-Markdown needs to match GitHub's rendering.

Affected pages include Downloads, Compiling, Running, the DFIR Timeline commands, Elastic Stack / Timesketch / Timeline Explorer guides, the Rust performance guide, and several Rules pages.

Builds clean with `mkdocs build --strict` (0 warnings); verified the binaries list now renders as a real `<ul>`.


YamatoSecurity (approved):
LGTM

PR #1794, opened 2026-06-28
docs: fix broken Hayabusa logo on non-English homepages

Shirofune-Security:
## Summary

The Hayabusa logo was broken on every homepage **except English**.

**Cause:** the hero used a raw HTML tag — ``. MkDocs only rewrites relative paths inside *Markdown* image syntax, not raw HTML attributes. So on translated homepages (served from `/ja/`, `/zh-TW/`, …) the `assets/logo.png` path resolved to `/<locale>/assets/logo.png`, which doesn't exist. (The same applied to the glightbox `href` it generated.)

**Fix:** use a Markdown image instead — `Hayabusa{ .hb-logo }`. MkDocs rewrites it per page location (e.g. `../assets/logo.png` under `/ja/`), so both the `` and the glightbox `href` resolve correctly. The `.hb-logo` class (and its styling) is preserved via `attr_list`.

Applied to all **15 homepages** (`index.md` + 14 locales). Builds clean with `mkdocs build --strict` (0 warnings); verified the logo resolves to the correct path on EN and JA.

### DIFF
--- a/website/docs/index.ar.md
+++ b/website/docs/index.ar.md
@@ -6,7 +6,7 @@ hide:
 
 <div class="hb-hero" markdown>
 
-<img class="hb-logo" alt="Hayabusa" src="assets/logo.png" />
+![Hayabusa](assets/logo.png){ .hb-logo }
 
 <p class="hb-tagline">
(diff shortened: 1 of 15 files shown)


YamatoSecurity (approved):
ok

PR #1798, opened 2026-06-30
fix: timeline::metrics nextest test failure (#1281)

Shirofune-Security:
## Fixes #1281

`cargo nextest run` fails on `timeline::metrics::tests::test_evt_logon_stats`:

```
thread 'timeline::metrics::tests::test_evt_logon_stats' panicked at src/detections/utils.rs:336:43:
called `Option::unwrap()` on a `None` value
```

### Root cause
The test calls `create_rec_info`, which reads the process-global `STORED_EKEY_ALIAS` — a `RwLock<Option<EventKeyAliasConfig>>` that starts as `None`:

```rust
let binding = STORED_EKEY_ALIAS.read().unwrap();
let eventkey_alias = binding.as_ref().unwrap();   // <- panics when None
```

Every *other* `create_rec_info` test initializes that global first (e.g. `timeline::timelines::tests::test_evt_logon_stats`, `rule::count`, `rule::mod` all do `*STORED_EKEY_ALIAS.write().unwrap() = Some(dummy_stored_static.eventkey_alias.clone());`). This one never did. Under the normal `cargo test` runner all tests share a single process, so some earlier test populates the global first and `test_evt_logon_stats` passes *by luck of ordering*. Under **nextest's process-per-test isolation** nothing sets it, so the `unwrap()` hits `None`.

### Fix
Initialize `STORED_EKEY_ALIAS` from the test's own dummy config before calling `create_rec_info`, exactly like the sibling tests:

```rust
*STORED_EKEY_ALIAS.write().unwrap() = Some(dummy_stored_static.eventkey_alias.clone());
```

The alias is loaded from the same `rules/config/eventkey_alias.txt` that every other test uses (`create_static_data` loads it unconditionally for all actions), so the written value is identical to what the other tests write — no new data race in the multi-threaded `cargo test` runner.

### Verification
nextest's prebuilt binary wouldn't run in my sandbox, so I reproduced its process-per-test isolation by running the single test in its own process (`<test-binary> --exact <name>`):

| check | before | after |
|---|---|---|
| `test_evt_logon_stats` in isolation (nextest-equivalent) | ❌ panics at `utils.rs:336` (`Option::unwrap()` on `None`) | ✅ passes |
| every `timeline::` test, one-per-process | metrics fails | ✅ all pass |
| full multi-threaded `cargo test --lib` (regression) | 423 pass | ✅ 423 pass, **25/25 runs** |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1799, opened 2026-06-30
test: fix remaining `cargo nextest run` failures in output-overwrite tests (#1281)

Shirofune-Security:
## Problem

Follow-up to #1798. On current `main`, `cargo nextest run` still fails the output-overwrite bin tests:

```
thread 'tests::test_same_file_output_logon_summary_csv_exit' panicked at src/main.rs:3513:9:
assertion `left == right` failed
  left: 214
 right: 0
```

(`test_same_file_output_computer_metrics_exit` and `test_same_file_output_metric_csv_exit` fail the same way.)

## Root cause

These tests come in **pairs**: a no-clobber `*_exit` variant that creates an empty output file and asserts the app leaves it untouched (size `0`), and a clobber variant that writes it. Each pair **shares one fixed output filename** (`overwrite.csv`, `overwrite.json`, `overwrite-metric*.csv`, `overwrite-computer-metrics.csv`).

I'd previously serialized them with an in-process `Mutex` (`FILE_OUTPUT_TEST_LOCK`, faa9179c). That works under `cargo test` (one process, many threads), but **nextest runs each test in its own process**, where a process-local mutex serializes nothing. So the clobber variant writes/deletes the very file the `*_exit` variant asserts is empty → `left: 214, right: 0`.

## Fix

Give every test a **unique output filename** so no two tests ever touch the same path, and drop the now-pointless lock. Correct under both runners, and removes the false cross-process safety the mutex implied. (39 insertions, 86 deletions — net simpler.)

## Verification

nextest's prebuilt binary won't execute in my sandbox, so I emulated its model exactly — each test spawned as its **own process**, all 10 run **concurrently** in a loop:

| | standard `cargo test --bin hayabusa` | concurrent processes (nextest model) |
|---|---|---|
| before | 17 pass | ❌ **61 failures / 400** (the `*_exit` tests) |
| after | ✅ 17 pass | ✅ **0 failures / 600** |

Relates to #1281

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3103,13 +3103,6 @@ mod tests {
         timeline::timelines::Timeline,
     };
 
-    /// Serializes the tests that exercise output-file overwrite behavior. They
-    /// share fixed output filenames (e.g. `overwrite-metric.csv`) between the
-    /// `*_exit` (no-clobber, expects the file untouched) and clobber variants,
-    /// so running them concurrently lets one test write the file the other
-    /// asserts is empty. Each such test holds this lock for its whole body.
-    static FILE_OUTPUT_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
-
     fn create_dummy_stored_static() -> StoredStatic {
         StoredStatic::create_static_data(Some(Config {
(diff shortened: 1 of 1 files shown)


YamatoSecurity (approved):
LGTM

PR #1802, opened 2026-07-02
ci(docs): auto-sync field-modifier count tables from hayabusa-rules

Shirofune-Security:
## What Changed

The **Supported Field Modifiers** page ([`rules/field-modifiers.md`](https://yamato-security.github.io/hayabusa/rules/field-modifiers/#hayabusa-supported-field-modifiers)) shows how often each field modifier / correlation rule is used across the Sigma and Hayabusa rulesets. Those counts are generated upstream in the hayabusa-rules repo (`doc/SupportedSigmaFieldModifiers.md`, regenerated whenever the rules change), and the docs copy had drifted stale. This adds automation to keep it current.

- **`website/scripts/sync_field_modifiers.py`** — fetches the upstream generated doc and replaces **only the data rows** of the two count tables in `field-modifiers.md` and all 13 translations. Localized headings, prose, and even localized table column headers (e.g. Turkish) are left untouched — only the numbers change. If the upstream table structure ever changes shape, the affected file is skipped and reported rather than corrupted.
- **`.github/workflows/sync-field-modifiers.yml`** — runs **weekly** (`cron: 17 6 * * 1`) and on demand (`workflow_dispatch`). When the counts change it commits the refreshed pages and dispatches the docs deploy so the site republishes.
- **Initial refresh** — ran the script once to fix the currently-stale counts (e.g. `contains` 3054 → 3069, `endswith` 3240 → 3250, `re` 186 → 188, `startswith` 532 → 535).

## Evidence

- The script's output for the English page now exactly matches the upstream tables (verified by a round-trip equality check).
- All 14 files produce clean, uniform diffs — only the 6 changed count rows per file (84 insertions / 84 deletions total); no headings, prose, or translated headers touched.
- `mkdocs build --strict` passes.
- Workflow YAML validated; `docs.yml` exposes `workflow_dispatch`, so the deploy dispatch works. Because pushes made with the built-in `GITHUB_TOKEN` don't trigger other workflows, the job explicitly dispatches `docs.yml` after committing.

### Notes

- This pushes directly to `main` (per the chosen auto-commit approach). If `main` has branch protection that blocks the `github-actions[bot]` push, switch the final step to open a PR instead (happy to provide that variant).
- The `cron` schedule can be tuned; weekly keeps churn low while staying reasonably fresh.


YamatoSecurity (approved):
LGTM

PR #1821, opened 2026-07-03
fix: null-UUID test-rule exemption in excluded/noisy rule counts compared the list-file path instead of the rule ID

### LINKED ISSUE #1814: [bug] Null-UUID test-rule exemption in excluded/noisy rule counts compares the list-file path instead of the rule ID, so it never applies
**Describe the bug**

In `ParseYaml::read_dir` and `count_rules` (src/yaml.rs), rules listed in `exclude_rules.txt` / `noisy_rules.txt` are counted into the "excluded"/"noisy" totals. Per the comment ("For test rules (ID: 000...0), exclude them from the excluded rule count."), rules with the null-UUID test ID `00000000-0000-0000-0000-000000000000` should be skipped in these counts. However, the guard compares `v`, which is the *value* returned from `RuleExclude::no_use_rule` — and that map stores rule-ID → list-file path (`no_use_rule.insert(v, filename.to_owned())` in src/filter.rs). Since a file path never equals the null UUID, the condition is always true and the exemption never takes effect. The nearby `utils::contains_str(v, "exclude_rule")` check confirms `v` is a path, not an ID. The guard has been ineffective since it was introduced in 9b03441 (for #602, whose report explicitly expected the test file not to be counted: "14 in the config file (+1 test file), so Excluded Rules: 14 should be displayed").

**Affected code**

`read_dir`: https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/yaml.rs#L348-L363

```rust
let rule_id = &yaml_doc["id"].as_str();
if rule_id.is_some() {
    if let Some(v) = exclude_ids
        .no_use_rule
        .get(&rule_id.unwrap_or(&String::default()).to_string())
    {
        let entry_key = if utils::contains_str(v, "exclude_rule") {
            "excluded"
        } else {
            "noisy"
        };
        // For test rules (ID: 000...0), exclude them from the excluded rule count.
        if v != "00000000-0000-0000-0000-000000000000" {
            let entry = self.rule_load_cnt.entry(entry_key.into()).or_insert(0);
            *entry += 1;
        }
```

Same pattern in `count_rules` (used by the wizard): https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/yaml.rs#L724-L736

```rust
        // For test rules (ID: 000...0), exclude them from the excluded rule count.
        if v != "00000000-0000-0000-0000-000000000000" {
```

The map is built with the rule ID as the key and the list-file path as the value: https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/filter.rs#L88-L95

```rust
for v in lines {
    let v = v.split('#').collect::<Vec<&str>>()[0].trim().to_string();
    if v.is_empty() || !configs::IDS_REGEX.is_match(&v) {
        // Ignore empty lines. Validate ID.
        continue;
    }
    self.no_use_rule.insert(v, filename.to_owned());
}
```

**Impact / example**

`rules/config/exclude_rules.txt` ships with the null UUID entry (`00000000-0000-0000-0000-000000000000 # TestFile`, [line 42](https://github.com/Yamato-Security/hayabusa-rules/blob/6dbc67d8446b719b40e3bcbf07b15dbb73dece1d/config/exclude_rules.txt#L42)). If a user has a test rule with `id: 00000000-0000-0000-0000-000000000000` in their rules directory, it is still counted in the "Excluded rules:" total printed at scan time (and in the wizard's per-status/level counts), contrary to the comment's stated intent of leaving test rules out of that count. The unit test `test_all_exclude_rules_file` (src/yaml.rs, asserts `rule_load_cnt["excluded"] == 5` for five null-UUID test rules) currently encodes this unintended behavior and only passes because the guard is dead.

**Possible fix**

Compare the rule's ID rather than the map value, e.g. `if rule_id != &Some("00000000-0000-0000-0000-000000000000")` (and equivalently in `count_rules`), or drop the dead check if counting test rules is actually acceptable. If fixed, `test_all_exclude_rules_file` would need its expected count adjusted (the five `exclude*.yml` fixtures all use the null UUID); the `noisy*.yml` fixtures use real UUIDs and are unaffected.

Shirofune-Security:
Closes #1814

## What Changed

`ParseYaml::read_dir()` and `count_rules()` exempt the null-UUID test rule (`00000000-0000-0000-0000-000000000000`, listed in `rules/config/exclude_rules.txt` as `# TestFile`) from the excluded/noisy rule counters. However, the guard compared `v` — the exclude/noisy **list-file path** (`no_use_rule` maps rule ID → file name, see `filter.rs::insert_ids`) — against the null UUID, so it was always true and the exemption never applied.

- Both guards now compare the **rule ID** (`rule_id.unwrap_or_default()`), matching the intent of the pre-existing comment (`// For test rules (ID: 000...0), exclude them from the excluded rule count.`).
- `test_all_exclude_rules_file` previously asserted the buggy behavior (`excluded == 5`; all five `test_files/rules/yaml/exclude*.yml` fixtures use the null-UUID ID). It now asserts the intended behavior: the count is exempted (`excluded == 0`) **and** the rules are still excluded from loading (no `exclude*.yml` in the loaded file list).
- `test_all_noisy_rules_file` is unaffected (`noisy == 5`) since the noisy fixtures use real UUIDs, so real-UUID counting stays covered.

Which rules get loaded or skipped is completely unchanged — only the counters shown in "Excluded rules" / "Noisy rules" and the wizard's per-level counts are affected, and only for rules with the null-UUID test ID. The shipped `rules/` repo currently contains no `.yml` with that ID, so production output is unchanged today; the fix matters when the test rule file is present.

This was flagged independently by Copilot on #1808 (which only documented the behavior, since that PR was comments-only): https://github.com/Yamato-Security/hayabusa/pull/1808#discussion_r3520149061

## Evidence

- `cargo fmt -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `RUST_TEST_THREADS=1 cargo test`: 442/442 pass, including the updated `test_all_exclude_rules_file` pinning both halves of the intended behavior (count exempted, rule still excluded).


YamatoSecurity (approved):
LGTM

PR #1824, opened 2026-07-03
fix: GeoIP lookup treated all public IPv6 addresses as Private

### LINKED ISSUE #1819: [bug] GeoIP lookup treats all public IPv6 addresses as Private (2000::/3 is in the private IP range list)
**Describe the bug**

`GeoIPSearch::check_in_private_ip_range()` includes `2000::/3` — the IPv6 global unicast block, i.e. essentially the entire publicly routable IPv6 address space — in its list of "private" ranges. Because `convert_ip_to_geo()` returns `Private🦅-🦅-` for any address that matches this list before performing a MaxMind lookup, every public IPv6 address is reported as `Private` and GeoIP enrichment effectively never works for IPv6 (only IPv4-mapped `::ffff:` addresses escape, since they are converted to IPv4 first). The unit test `test_check_in_private_range_v6` currently asserts this behavior for a global unicast address. As a minor side note, `FD00::/8` in the same list is redundant because it is a subset of `FC00::/7`.

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/geoip_search.rs#L67-L76

```rust
        } else {
            vec![
                IpCidr::from_str("::/128").unwrap(),    // IPv6 Unspecified
                IpCidr::from_str("2000::/3").unwrap(),  // IPv6 Global Unicast
                IpCidr::from_str("FE80::/10").unwrap(), // IPv6 Link Local Unicast
                IpCidr::from_str("FC00::/7").unwrap(),  // IPv6 Unique Local Address
                IpCidr::from_str("FD00::/8").unwrap(),  // IPv6 Unique Local Address
                IpCidr::from_str("FF00::/8").unwrap(),  // IPv6 Multicast Address
            ]
        };
```

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/geoip_search.rs#L107-L109

```rust
        if self.check_in_private_ip_range(&addr) {
            return Ok("Private🦅-🦅-".to_string());
        }
```

The unit test asserts the current behavior, so it would need updating along with a fix:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/geoip_search.rs#L306-L308

```rust
        let global_unicast = geo_ip.convert_ip_to_geo("2001:1234:abcd:1234::1");
        assert!(global_unicast.is_ok());
        assert_eq!("Private🦅-🦅-", global_unicast.unwrap());
```

**Impact / example**

When running with `-G <mmdb dir>` (the `SrcASN`/`SrcCountry`/`SrcCity` and `TgtASN`/`TgtCountry`/`TgtCity` columns are appended to the output profile automatically), any event whose source/target address is a public IPv6 address — for example `2606:4700::6810:84e5` (Cloudflare) or the `2001:1234:abcd:1234::1` used in the test — is output as `SrcASN: Private`, `SrcCountry: -`, `SrcCity: -` instead of the actual ASN/country/city (the `Private🦅-🦅-` string is split into the three fields in `src/detections/detection.rs`). In other words, GeoIP lookups silently never happen for native IPv6 traffic, which matters increasingly for RDP/logon events on IPv6-enabled networks.

For context, the private-range check was added in commit 9f26ca0c (PR #896), where the review request was specifically to output `Private` for private IP addresses, which suggests that including the global unicast block in the list was unintentional.

**Possible fix**

Remove `2000::/3` from the list (and optionally drop the redundant `FD00::/8`), leaving only genuinely non-global ranges: `::/128`, `FE80::/10`, `FC00::/7`, `FF00::/8`. The assertion in `test_check_in_private_range_v6` would then be updated to expect a real lookup result (or a distinct non-private test address) for global unicast input.

Shirofune-Security:
## What

Fixes #1819.

`GeoIPSearch::check_in_private_ip_range()` included `2000::/3` — the entire IPv6 global unicast (publicly routable) space — in its private range list, so `convert_ip_to_geo()` returned `Private🦅-🦅-` for every public IPv6 address before ever consulting the MaxMind databases. GeoIP enrichment therefore silently never worked for native IPv6 traffic (only IPv4-mapped `::ffff:` addresses escaped, since they are converted to IPv4 first).

Changes:
- Removed `2000::/3` from the IPv6 private range list, along with the redundant `FD00::/8` (a subset of `FC00::/7`). The list now contains only genuinely non-global ranges: `::/128`, `FE80::/10`, `FC00::/7`, `FF00::/8`. (IPv6 loopback `::1` is already handled separately via `is_loopback()`.)
- Updated `test_check_in_private_range_v6`, which previously asserted the buggy behavior: a global unicast address (`2001:218::1`, Japan in the MaxMind test data) must now return a real lookup result, and added unspecified/unique-local/multicast cases that must stay `Private`.
- De-flaked the geoip tests: they run in parallel and share the global `IP_MAP` cache, and the `IP_MAP.clear()` calls plus the dummy-cache test reusing the same IP as the real-lookup test could race (the real-lookup test could observe the dummy value, or the dummy could be cleared mid-test). Each test now uses addresses no other test touches, and the clears are gone. Verified stable across 10 consecutive runs.

## Verification

Ran `csv-timeline -J -G test_files/mmdb` end-to-end on crafted 4624 events with a public IPv6, a link-local IPv6, and an RFC1918 IPv4 source address.

Before (current main) — all three, including the public IPv6, reported as Private:
```
SrcIP: 2001:218::1  → SrcASN: Private, SrcCountry: -, SrcCity: -
SrcIP: fe80::1234   → SrcASN: Private, SrcCountry: -, SrcCity: -
SrcIP: 192.168.1.10 → SrcASN: Private, SrcCountry: -, SrcCity: -
```

After — the public IPv6 gets a real lookup; non-global addresses still report Private:
```
SrcIP: 2001:218::1  → SrcCountry: Japan (no ASN/City entry in the test data)
SrcIP: fe80::1234   → SrcASN: Private, SrcCountry: -, SrcCity: -
SrcIP: 192.168.1.10 → SrcASN: Private, SrcCountry: -, SrcCity: -
```

`cargo test`, `cargo fmt --check`, and `cargo clippy` are clean.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -15,6 +15,7 @@
 - null-UUIDのテストルールを除外ルール・ノイジールールの件数から除くための判定が、ルールIDではなく除外リストファイルのパスと比較していたため機能しておらず、テストルールが`Excluded rules`の件数を水増ししていた問題を修正した。 (#1821) (@YamatoSecurity)
 - Results Summary（およびHTMLレポート）のユニーク検知のパーセンテージがレベル間で鏡写しに入れ替わっていた問題を修正した。逆順ループのインデックスでパーセンテージを計算していたため、例えば`critical`の行に`informational`のパーセンテージが表示されていた（正しかったのは中央の`medium`のみ）。 (#1812) (@YamatoSecurity)
 - `logon-summary`コマンドで、RDSゲートウェイのログオン（`Microsoft-Windows-TerminalServices-Gateway/Operational`のEID 302）のTarget Domain列が常に`-`になっていた問題を修正した。`dst_domain`の抽出がイベントキーエイリアス`RdsGtwUsername`をスペルミスの`RdsGtwUserName`で参照していたため、`DOMAIN\user`形式の値からドメインが取得されていなかった。 (#1809) (@YamatoSecurity)
+- GeoIP機能がすべてのパブリックIPv6アドレスを`Private`として扱い、GeoIP検索を行っていなかった問題を修正した。IPv6のプライベート範囲リストにグローバルユニキャスト空間全体である`2000::/3`が含まれていたことが原因。あわせて冗長な`FD00::/8`（`FC00::/7`に包含される）も削除し、パブリックIPv6アドレスにASN・国・都市の情報が付与されるようにした。 (#1819) (@YamatoSecurity)
 
 ## 3.9.0 [2026/04/29]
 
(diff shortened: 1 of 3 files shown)


YamatoSecurity (approved):
LGTM

Issue #1807, opened 2026-07-03
Translate remaining Japanese code comments to English and improve comment quality


Shirofune-Security:
## Summary

Hayabusa's source still contains a number of Japanese-language code comments, several half-translated comments left over from earlier translation passes (an English sentence with a Japanese fragment still attached), and quite a few comments that are too terse or whose English is hard to follow. There are also some misspelled identifiers (e.g. `AggegationConditionCompiler`, `CONTROL_CHAT_REPLACE_MAP`, `get_childs`, `scanable_rule_exists`, `rap_checkpoint`) that hurt readability and greppability.

Since Hayabusa has an international user and contributor base, it would be helpful if all code comments were consistent, accurate English, and if obvious identifier misspellings were fixed.

## Proposal

A comments/identifiers-only cleanup PR that:

1. Translates all remaining Japanese comments in `src/**/*.rs` into English, preserving the full original meaning (including author TODO/self-critique notes kept as refactoring reminders).
2. Repairs the half-translated comment artifacts (e.g. in `matchers.rs`).
3. Cleans up existing English comments (spelling, grammar, clarity) and rewrites comments that no longer match what the code actually does.
4. Adds comments/doc comments where the logic is hard to follow (rule engine internals, wildcard→regex conversion, count/correlation logic, output profiles, etc.).
5. Fixes misspelled identifiers (internal names only — no CLI-visible strings, no serde/output field names, no string literals).

Strictly **no behavior changes**: string literals stay byte-for-byte identical (including the Japanese strings used as UTF-8/UTF-16 test data and user-facing messages), and the full test suite must pass unchanged.

## Out of scope (noted for follow-up)

Typos inside **user-facing string literals** were deliberately left untouched so this PR stays reviewable as a pure comment/identifier change, e.g.:

- `"An unknown error occured."` (`matchers.rs`) → "occurred"
- `"- Selected deteciton rule set: ..."` (`main.rs`, HTML report) → "detection"
- `"Successed submodule update"` (`options/update.rs`) → "Succeeded"
- the test config file name `test_files/config/regex/detectlist_suspicous_services.txt` → "suspicious"

<details>
<summary>Possible bugs noticed while reading the code for this effort (not addressed in the PR; now filed as individual issues — see links below)</summary>

While reading every file carefully enough to write accurate comments, we noticed a few things that look like real bugs. They are **not** touched by the translation PR. Each item was re-verified against `main` and has been filed as its own issue (linked below):

[...]

PR #1826, opened 2026-07-03
fix: anchor wildcard field-value regexes to prevent substring false positives (#1815)

### LINKED ISSUE #1815: [bug] Wildcard patterns that fall back to regex matching are unanchored, so they substring-match instead of full-value matching
**Describe the bug**

Field patterns that cannot be converted to a `FastMatch` fall back to a regex built by `PipeElement::pipe_pattern_wildcard()`. That regex is never anchored with `^`/`$`, and despite its name `DefaultMatcher::is_regex_fullmatch()` uses `Regex::is_match()`, which searches for a match *anywhere* in the value. As a result, these patterns match field values that have extra leading/trailing characters, deviating from Sigma full-value match semantics and from Hayabusa's own fast-match paths (`FastMatch::Exact` requires equal length, `StartsWith`/`EndsWith` are anchored). The fallback is taken whenever `convert_to_fast_match()` returns `None` — i.e. the pattern contains `?`, ends with `\\\*`, is non-ASCII and contains `*`, or contains `*` in any other shape not recognized as a plain prefix/suffix/contains pattern (e.g. a mid-string `*`) — and also when a `StartsWith`/`EndsWith` fast match meets a non-ASCII event value (the `*_ignore_case` helpers return `None` there and control falls through to the regex).

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/rule/matchers.rs#L216-L219

```rust
    /// Determines whether this matcher matches the given regex.
    fn is_regex_fullmatch(&self, value: &str) -> bool {
        self.re.as_ref().unwrap().iter().any(|x| x.is_match(value))
    }
```

`Regex::is_match()` returns true if the regex matches anywhere in the haystack, so this is not a "fullmatch" unless the regex itself is anchored — but the wildcard-to-regex conversion never adds anchors:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/rule/matchers.rs#L902-L904

```rust
        // sigma wildcards are case insensitive.
        // Therefore, prepend a symbol indicating case insensitivity to the regex.
        "(?i)".to_string() + &ret
```

The regex fallback is reached from `is_match()`:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/rule/matchers.rs#L599-L600

```rust
        // If it could not be converted to character count/starts_with/ends_with search, compare using regex match.
        self.is_regex_fullmatch(event_value_str)
```

The pattern shapes that skip fast matching (and therefore hit the unanchored regex) are decided here:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/rule/fast_match.rs#L53-L58

```rust
    if utils::contains_str(s, "?")
        || s.ends_with(r"\\\*")
        || (!s.is_ascii() && utils::contains_str(s, "*"))
    {
        // Patterns that cannot be converted to fast matching use regex match only.
        return None;
```

(plus the catch-all for other `*` placements, e.g. a mid-string `*`, at [fast_match.rs#L84-L87](https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/rule/fast_match.rs#L84-L87)).

**Impact / example**

Rules with these pattern shapes produce false positives. For example:

```yaml
detection:
    selection:
        Channel: Sec?rity
```

Per Sigma semantics, `Sec?rity` should only match 8-character values such as `Security` / `Sec1rity`. In Hayabusa it becomes the unanchored regex `(?i)Sec.rity`, so a record with `"Channel": "MySec1rityLog"` also matches. The behavior is internally inconsistent: the wildcard-free pattern `Channel: Security` correctly does **not** match `"MySecurityLog"` (it takes the `FastMatch::Exact` path, which compares lengths). Similarly, `CommandLine: net*user` (mid-string `*`) matches `"mynetXuserZ"`, and a non-ASCII pattern like `Channel: ホスト*` matches values that merely *contain* `ホスト` instead of starting with it. Among the regex-based paths, only the keyword/grep search (empty `key_list`, matchers.rs L585-L592) and the `|re` modifier (see `test_detect_regex_partial_match`) are intentionally unanchored.

**Possible fix**

Anchor the regex produced by `pipe_pattern_wildcard()` (e.g. `format!("(?i)^{ret}$")`). The `|re`/`|reignorecase`/etc. modifiers are unaffected by this because `PipeElement::Wildcard` is not pushed for them (matchers.rs L485-L496), but the keyword/grep search path (empty `key_list`) also uses the wildcard-built regexes, so it would need to keep an unanchored variant. Alternatively, `is_regex_fullmatch()` could verify that the match spans the entire value — but `|re` field matches also reach `is_regex_fullmatch()` and are intentionally partial (see `test_detect_regex_partial_match`), so that path would need a carve-out.

Shirofune-Security:
## What

Fixes #1815.

Field patterns that cannot be converted to a `FastMatch` fall back to a regex built by
`PipeElement::pipe_pattern_wildcard()`. That regex is never anchored, and despite its name
`DefaultMatcher::is_regex_fullmatch()` uses `Regex::is_match()`, which matches *anywhere* in the
value. So wildcard field patterns **substring-matched** instead of full-value matching, deviating
from Sigma semantics and from Hayabusa's own fast-match paths (`FastMatch::Exact` compares length;
`StartsWith`/`EndsWith` are anchored), producing false positives.

The regex fallback is taken when `convert_to_fast_match()` returns `None` — a pattern with `?`, a
mid-string `*`, a trailing `\\\*`, or non-ASCII text plus `*` — and also when a `|startswith` /
`|endswith` fast match meets a non-ASCII event value (the `*_ignore_case` helpers return `None` and
control falls through to the regex).

**Examples that wrongly matched before this change**

| Rule pattern | Value | Before | After |
|---|---|---|---|
| `Channel: Sec?rity` | `MySec1rityLog` | ✅ matched | ❌ no match |
| `CommandLine: net*user` | `mynetXuserZ` | ✅ matched | ❌ no match |
| `Channel: ホスト*` | `Myホストログ` | ✅ matched | ❌ no match |

## How

Anchor the compiled wildcard regex with `^(?:…)$` for field matches, in `DefaultMatcher::init()`.

Two regex paths intentionally keep **substring** semantics and are left unanchored:
- **Keyword (grep) searches** — an empty key list; matched against the whole-record string.
- **The keyless `|all` modifier** — its leaf is matched against the whole-record string too
  (`selectionnodes.rs`), so `key_list[0] == "|all"` is excluded from anchoring as well.

`|re` / `|reignorecase` / `|remultiline` / `|resingleline` are unaffected because
`PipeElement::Wildcard` is not applied to them (they stay intentionally partial;
`test_detect_regex_partial_match` still passes).

## Tests

Adds regression tests for `?`, mid-string `*`, and non-ASCII `*` field patterns (full-value match
vs. no substring match), the `|startswith` non-ASCII regex fallback, and guards that keyword/grep
and `|all` searches keep their substring semantics.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and
`RUST_TEST_THREADS=1 cargo test` all pass.

## Note

This changes detection behavior for rules that relied (unintentionally) on the substring match.
That behavior was already inconsistent — a wildcard-free `Channel: Security` correctly does **not**
match `MySecurityLog` via the `FastMatch::Exact` path — so this aligns the wildcard paths with the
documented Sigma full-value semantics. Found while reading the source for #1807.


YamatoSecurity (approved):
LGTM

PR #1827, opened 2026-07-04
fix: render General Overview details dropped by a misspelled HTML-report section key (#1818)

### LINKED ISSUE #1818: [bug] "Analyzed event files" / total file size lines never appear in the HTML report due to a misspelled section key
**Describe the bug**

When the HTML report is enabled, `analysis_files()` registers the analyzed-file count, total file size, selected detection rule set, and excluded tags under the section key `"General Overview #{general_overview}"` — with the `#` outside the braces. Every other call site in the codebase uses `"General Overview {#general_overview}"`, which is the key registered in `section_order` by `get_init_md_data_map()`. Because `add_md_data()` stores data under whatever key it is given and `create_html()` only renders sections whose names appear in `section_order`, everything added under the misspelled key ends up in an orphan map entry that is never rendered — the data is silently dropped from the report.

**Affected code**

The misspelled key (all other call sites use `{#general_overview}`):

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/main.rs#L1788-L1813

```rust
        if stored_static.html_report_flag {
            let mut output_data = Nested::<String>::new();
            let mut html_report_data = Nested::<String>::from_iter(vec![
                format!(
                    "- Analyzed event files: {}",
                    evtx_files.len().to_formatted_string(&Locale::en)
                ),
                format!("- {total_size}"),
            ]);
            // ... (adds "Selected deteciton rule set: ..." and "Excluded tags: ..." lines)
            output_data.extend(html_report_data.iter());
            htmlreport::add_md_data("General Overview #{general_overview}", output_data);
        }
```

The canonical section keys and the render loop that drops any other key:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/htmlreport.rs#L92-L103

```rust
    section_order.extend(vec![
        "General Overview {#general_overview}",
        "Results Summary {#results_summary}",
    ]);
```

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/htmlreport.rs#L54-L64

```rust
        for section_name in self.section_order.iter() {
            if let Some(v) = self.md_datas.get(section_name) {
                md_data.push(format!("## {}\n", &section_name));
```

`add_md_data()` (https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/options/htmlreport.rs#L106-L115) uses `entry(section_name).or_insert(...)`, so the misspelled key simply creates a new `md_datas` entry that the loop above never reads.

**Impact / example**

Running e.g. `hayabusa csv-timeline -d ./logs -o out.csv -H report.html`, the "General Overview" section of `report.html` contains the command line, start time, output profile, elapsed time, etc. (added elsewhere with the correct key), but is missing the lines produced in this block:

- `Analyzed event files: 581`
- the total file size line
- `Selected detection rule set: ...` (the wizard selection)
- `Excluded tags: ...` (when tags were excluded)

The unit tests in `src/options/htmlreport.rs` (e.g. L231–L232 expect `- Analyzed event files: 581` and `- Total file size: 148.5 MB`) show these lines are intended to be part of General Overview. The typo was introduced in commit 974ec45c ("added id attribute to html header tag #689", Sep 2022), when the `{#id}` heading attributes were added: the other two call sites in that commit were changed to `{#general_overview}` while this one got `#{general_overview}`. Current releases are affected. (A later cleanup, 74fe5b3a, touched the same line, so `git blame` points there.)

**Possible fix**

Change the key at `src/main.rs:1812` to `"General Overview {#general_overview}"`, or introduce a shared constant for the section names so the key cannot drift again. Note that once this block actually renders, a few latent nits in it become user-visible and may be worth fixing in the same PR: the size line is formatted as `format!("- {total_size}")` without the `Total file size: ` label the tests expect, `deteciton` is a typo for `detection`, and the wizard line comes out double-labeled because `status_append_output` (src/main.rs:1620-1623) already starts with `- selected detection rule sets: `.

Shirofune-Security:
Closes #1818

## Problem

When the HTML report is enabled, `analysis_files()` registered the analyzed-file count, total file size, selected detection rule set, and excluded tags under the section key `"General Overview #{general_overview}"` — with the `#` **outside** the braces. Every other call site, and `section_order` in `get_init_md_data_map()`, uses `"General Overview {#general_overview}"`.

`create_html()` only renders sections whose key is present in `section_order`, so the misspelled key created an orphan `md_datas` entry that was never read — those four lines were silently dropped from the report (introduced by the `{#id}`-attribute change in commit `974ec45c`, Sep 2022).

## Fix

- **Root cause**: introduced shared constants `GENERAL_OVERVIEW_SECTION` / `RESULTS_SUMMARY_SECTION` in `options::htmlreport` and use them at every call site (`section_order` builder + the four `main.rs` sites), so the section keys have a single source of truth and can't drift again.
- **Latent issues that became user-visible once the block renders** (called out in the issue):
  - The total-size line was `format!("- {total_size}")` → now `"- Total file size: {total_size}"`, matching the other lines and the `test_create_html` expectation (`- Total file size: 148.5 MB`).
  - Fixed the `deteciton` → `detection` typo in the rule-set label.
  - Fixed the double-labeled rule-set line: `status_append_output` already began with `- selected detection rule sets: `, and the report block wrapped it again as `- Selected deteciton rule set: {…}`. `status_append_output` now holds just the selected set name and the report block supplies the single `- Selected detection rule set: ` prefix.

## Result

The General Overview section of the HTML report now includes the analyzed-file count, total file size, selected rule set, and excluded tags (when set), each with a correct single label.

## Tests

- Added `test_general_overview_section_key_is_rendered` (regression for #1818): data registered under `GENERAL_OVERVIEW_SECTION` is rendered, while data under the old misspelled key is not.
- `cargo fmt -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `RUST_TEST_THREADS=1 cargo test`: **456/456 pass** (439 lib + 17 bin) ✅

Found while reading the source for #1807.


YamatoSecurity (approved):
LGTM!

PR #1834, opened 2026-07-04
fix: three latent error-handling / dead-code issues (#1831, #1832, #1833)

### LINKED ISSUE #1831: [bug] read_encoded_file silently ignores read errors (and panics on invalid UTF-8)
## Summary

`ParseYaml::read_encoded_file()` in `src/yaml.rs` silently discards the result of `read_to_end`, so an I/O error while reading an obfuscated rule file (`encoded_rules.yml`) is treated as a valid (truncated or empty) file instead of being reported. The same function also panics on invalid UTF-8 rather than returning an error.

## Location

`src/yaml.rs`, `read_encoded_file()` (around lines 101–111 on `main`):

```rust
fn read_encoded_file(path: &PathBuf) -> Result<String, String> {
    let mut fr = fs::File::open(path)
        .map(BufReader::new)
        .map_err(|e| e.to_string())?;
    let mut encrypted_content = Vec::new();
    let _ = fr.read_to_end(&mut encrypted_content);          // (1) result discarded
    let decode_content = encrypted_content.iter().map(|&b| b ^ 0xAA).collect();
    let decode_string = String::from_utf8(decode_content).expect("Invalid UTF-8 sequence"); // (2) panics
    Ok(decode_string)
}
```

## Problems

1. **Ignored read error.** `let _ = fr.read_to_end(...)` throws away the `io::Result`. If the read fails part-way (disk error, etc.), whatever bytes were read so far are XOR-decoded and returned as `Ok(...)`, so a real I/O failure is silently turned into a truncated/empty rule set. The function already returns `Result<_, String>` and its sibling `read_file()` handles this correctly:

   ```rust
   fr.read_to_string(&mut file_content).map_err(|e| e.to_string())?;
   ```

2. **Panic on invalid UTF-8.** `String::from_utf8(decode_content).expect("Invalid UTF-8 sequence")` aborts the whole process if the decoded bytes aren't valid UTF-8 (e.g. a corrupted or wrongly-encoded `encoded_rules.yml`), instead of surfacing an `Err`.

## Suggested fix

```rust
    fr.read_to_end(&mut encrypted_content).map_err(|e| e.to_string())?;
    let decode_content: Vec<u8> = encrypted_content.iter().map(|&b| b ^ 0xAA).collect();
    let decode_string = String::from_utf8(decode_content).map_err(|e| e.to_string())?;
    Ok(decode_string)
```

## Notes

Surfaced during the variable-naming review in #1830. Impact is low-probability (needs a mid-read I/O error or a corrupt encoded file) but it masks a genuine failure as success. Happy to open a PR if you'd like this fixed.

### LINKED ISSUE #1832: Dead branch in _convert_valid_json_str: addition_header is always empty
## Summary

In `_convert_valid_json_str()` (`src/afterfact.rs`), the `addition_header` binding inside the `starts_with('"')` branch is computed from a condition that can never be true, so it is always `""`. It is dead code, and it may hint that the intended leading-quote handling never runs.

## Location

`src/afterfact.rs`, `_convert_valid_json_str()` (around line 1909 on `main`):

```rust
    } else if con_cal.starts_with('"') {
        let addition_header = if !con_cal.starts_with('"') { "\"" } else { "" };
        //                        ^^^^^^^^^^^^^^^^^^^^^^^^^ always false in this branch
        let addition_quote = if !con_cal.ends_with('"') && concat_flag {
            "\""
        } else if !con_cal.ends_with('"') {
            "\\\""
        } else {
            ""
        };
        [
            addition_header,   // always ""
            &con_cal.replace('🛂', "\\").replace('\\', "\\\\").replace('"', "\\\""),
            addition_quote,
        ]
        .join("")
    } else {
        ...
    }
```

## Problem

This block is only entered when `con_cal.starts_with('"')` is **true**, so `!con_cal.starts_with('"')` is **always false**, and `addition_header` is **always `""`**. It contributes nothing to the joined result.

At minimum this is dead code that should be removed for clarity. It may also indicate a logic slip: the "prepend an opening quote when the value doesn't already start with one" intent can never fire here, so it's worth confirming that the JSON escaping is actually correct for values that start with `"` versus those that don't (the non-`"` case is handled by the final `else` branch, which adds neither a header nor a trailing quote).

## Suggested direction

If the current output is correct, drop `addition_header` and inline the two remaining pieces. If a leading quote was actually intended for some inputs, move that logic to where it can run. I didn't change this in #1830 (a rename-only PR) to keep that diff focused.

## Notes

Surfaced during the variable-naming review in #1830. Happy to open a PR once the intended behavior is confirmed.

### LINKED ISSUE #1833: Dead branch and needless Vec allocation in get_time_offset (configs.rs)
## Summary

In the `get_time_offset` closure in `src/detections/configs.rs`, the relative-offset digit extraction has a branch that can never run and allocates a `Vec` on every iteration just to read its last element.

## Location

`src/detections/configs.rs`, `get_time_offset` (around line 2198 on `main`):

```rust
let mut timekey_splitter = timeline_offset.split(*key);
let mix_check = timekey_splitter.next();
let mixed_checker: Vec<&str> = mix_check.unwrap_or_default().split(timekey).collect();
let target_num = if mixed_checker.is_empty() {
    mix_check.unwrap()                              // (1) dead branch
} else {
    mixed_checker[mixed_checker.len() - 1]          // (2) via a throwaway Vec
};
```

## Problem

- `str::split(...)` always yields **at least one** element (even for an empty string it yields `[""]`), so `mixed_checker` is never empty. The `if mixed_checker.is_empty()` branch — and its `mix_check.unwrap()` — is **dead code**.
- Collecting into a `Vec<&str>` only to index its last element is an unnecessary allocation, done once per unit (`y M d h m s`) per `get_time_offset` call.

Behavior is otherwise correct; this is a readability + minor-efficiency cleanup, not a functional bug.

## Suggested fix

Use `.split(...).last()` (or `.next_back()`) directly, which is guaranteed `Some` and avoids the allocation and the dead branch:

```rust
let before_unit = timeline_offset.split(*key).next().unwrap_or_default();
let target_num = before_unit.split(timekey).last().unwrap_or_default();
```

## Notes

Surfaced during the variable-naming review in #1830 (there the two locals are renamed to `before_unit` / `unit_digit_segments`, which made the dead branch easier to spot). Not changed in that PR since it is rename-only. Happy to open a PR.

Shirofune-Security:
Fixes three latent issues found during the #1830 review and tracked as separate issues.

### #1831 — `read_encoded_file` swallows errors (`src/yaml.rs`)
`let _ = file_reader.read_to_end(...)` discarded the `io::Result`, so a mid-read I/O error was silently returned as a truncated/empty rule file; and `String::from_utf8(...).expect(...)` **panicked the whole process** on a corrupt `encoded_rules.yml`. Both are now propagated through the function's existing `Result<_, String>` (matching the sibling `read_file`). Added a regression test for the invalid-UTF-8 case.

### #1832 — dead `addition_header` branch (`src/afterfact.rs`)
In `_convert_valid_json_str`, inside the `joined_value.starts_with('"')` branch, `addition_header` was `if !joined_value.starts_with('"') { … }` — always false there, so it was always `""`. Removed it. No behavior change; added a regression guard test for quote-prefixed values.

### #1833 — dead branch + needless allocation (`src/detections/configs.rs`)
In `get_time_offset`, `str::split` always yields ≥1 element, so `unit_digit_segments.is_empty()` was never true (dead branch) and the code collected into a `Vec` only to read its last element. Now uses `split(...).next_back()` on the `&str` directly. Behavior unchanged (covered by the existing time-offset tests).

## Safety / testing

No struct is serde-serialized, and #1832/#1833 are behavior-preserving (the removed branches were provably dead). `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `RUST_TEST_THREADS=1 cargo test` (478 lib + 17 integration tests, including 2 new regression tests) all pass.

Closes #1831
Closes #1832
Closes #1833


YamatoSecurity (approved):
LGTM

PR #1837, opened 2026-07-04
ci: stop bundling README/CHANGELOG PDFs in release packages

Shirofune-Security:
Removes PDF generation from the release workflow. Documentation now lives on the website, so the release packages no longer need PDF copies of the markdown files.

## Also fixes the 3.10.0 release failure
The macOS `Create PDF` step ran `md-to-pdf ./*.md` then `mv ./README-Japanese.pdf …`. Since the README became a landing page (`4a6ba73b`), the Japanese README was archived to `OLD-README-Japanese.md`, so `README-Japanese.pdf` is never produced and the `mv` failed with `No such file or directory` → the whole macOS upload job failed. (Example: [run 28707396161](https://github.com/Yamato-Security/hayabusa/actions/runs/28707396161/job/85136732702).)

## Change
Removed three steps from the macOS `upload` job:
- **Setup node** (only present to `npm i -g md-to-pdf`)
- **Create PDF** (`md-to-pdf` + the `mv` renames)
- **Upload Document Artifacts** (uploaded `./*.pdf` as `hayabusa-documents`)

No downstream breakage: the `upload-all-platforms` / `all-packages-zip` jobs download with `pattern: hayabusa-*`, which simply matches the binary artifacts now (one fewer artifact, no error). The release just ships without PDFs.

> Note: this was also committed on the `finalize-3.10.0` branch, but #1836 was merged before it landed, so it's not in `main` — hence this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -209,28 +209,6 @@ jobs:
           path: release-binaries-encoded-rules/*
           include-hidden-files: true
 
-      - name: Setup node
-        if: matrix.info.target == 'aarch64-apple-darwin'
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
-        with:
-          node-version: 20
-
-      - name: Create PDF
-        if: matrix.info.target == 'aarch64-apple-darwin'
-        run: |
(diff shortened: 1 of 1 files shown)


YamatoSecurity (approved):
LGTM

Issue #1810, opened 2026-07-03
[bug] temporal_ordered correlation does not enforce event order beyond the first rule (no-op `last_base = base;` in detect_within_timeframe)


Shirofune-Security:
**Describe the bug**

In `Detection::detect_within_timeframe()` (src/detections/detection.rs), the variable `last_base` is initialized to `base` before the inner loop and then re-assigned `last_base = base;` at the end of each iteration. Since `base` never changes inside the inner loop, this assignment is a no-op and `last_base` always stays equal to the first rule's result. As a result, for `temporal_ordered` correlation rules, each referenced rule 2..n is only checked to occur at-or-after the **first** rule's result (within the timeframe) — the relative order among rules 2..n themselves is never enforced. Per the Sigma correlations specification, `temporal_ordered` should require the events to occur in the order the rules are listed. The variable name and the placement of the update suggest the intent was to advance `last_base` to the timestamp of the record matched for the previous rule.

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/detection.rs#L212-L237

```rust
            for base in base_records {
                let mut found = false;
                let mut last_base = base;
                for id in ids.iter().skip(1) {
                    found = false;
                    if let Some(target_records) = temporal_ref_all_results.get(id.as_str()) {
                        if temporal_ordered {
                            found = target_records.iter().any(|t| {
                                (t.start_timedate >= last_base.start_timedate)
                                    && (t.start_timedate <= last_base.start_timedate + timeframe)
                            });
                        } else {
                            // ...
                        }
                        if !found {
                            break;
                        }
                        last_base = base; // no-op: last_base already equals base
                    }
                }
                if found {
                    result.push(base.clone());
                }
```

(Introduced in d95d33db, "feat: add support for temporal_ordered", for #1447.)

**Impact / example**

A `temporal_ordered` rule behaves almost identically to `temporal`: the only ordering actually enforced is that rules 2..n occur at-or-after rule 1. For example, with:

```yaml
correlation:
  type: temporal_ordered
  rules:
    - rule_a
    - rule_b
    - rule_c
  group-by:
    - Computer
  timespan: 10m
```

[...]

PR #1839, opened 2026-07-04
fix: enforce group-by equality in temporal correlations

Shirofune-Security:
Related to #1810 (follow-up surfaced while reviewing #1838, the ordering fix for that issue).

## Summary

`temporal`/`temporal_ordered` correlations did not enforce the rule's `group-by` value when combining referenced-rule matches, so events from **different groups** (e.g. different `Computer`s) could be correlated together and raise false-positive alerts.

This was surfaced by Copilot's review of #1838 (the `temporal_ordered` ordering fix for #1810). It is a **separate, pre-existing** limitation — the combine step in `Detection::detect_within_timeframe` only compared timestamps — and was explicitly documented as known behavior in `parse_temporal_rules` ("group equality is not checked when combining"). This PR closes that gap independently of #1838.

## Details

Each referenced rule is aggregated per group-by value, and `AggResult.key` holds that value (built by `create_count_key` from the temporal rule's `group-by` fields, so it is formatted identically across every referenced rule). The combine step now requires each referenced-rule match to share the base match's `key`, so only events from the same group are correlated. When `group-by` is effectively absent the key is a constant (`"_"`), so the check is a harmless no-op in that case.

The stale "group equality is not checked" note in `parse_temporal_rules` is updated to match the new behavior.

## Testing

- Added `test_detect_within_timeframe_enforces_group_by`, covering, for both ordered and unordered correlations:
  - referenced match in a **different** group → does **not** correlate
  - referenced match in the **same** group (alongside a different-group decoy) → correlates, returning the correct group
- `cargo test --lib detections::detection -- --skip geoip` passes (skipped geoip tests need `update-rules` config data, unrelated).
- `cargo fmt --check` and `cargo clippy --lib` clean.

## Note on #1838

This PR and #1838 both touch `detect_within_timeframe` (they are intentionally independent). Whichever merges first, the other needs a trivial rebase — the `group-by` check simply moves into the reworked ordered-matching iterator introduced by #1838.

## Changelog

Added entries to `CHANGELOG.md` and `CHANGELOG-Japanese.md`.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -5,7 +5,7 @@
 **バグ修正:**
 
 - `update-rules` 実行後に表示される文法的に誤ったメッセージ「Successed submodule update」を「Submodule update succeeded」に修正した。 (#1840) (@YamatoSecurity)
-- `temporal`/`temporal_ordered` 相関が、参照ルールのマッチを結合する際にルールの `group-by` 値を検証していなかった問題を修正した。各参照ルールは group-by 値ごとに集計されるものの、タイムスタンプがタイムフレームに収まってさえいれば異なるグループ（例: 異なる Computer）のマッチ同士が相関してしまい、誤検知を生んでいた。参照ルールのマッチは、ベースとなるマッチと同じ group-by 値を持つことを要求するようにした。 (@Shirofune-Security)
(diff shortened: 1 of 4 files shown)


YamatoSecurity:
@fukusuket Could you check this one too? It's a different bug but related to temporal correlation

Shirofune-Security:
Superseded by #1841, which combines this fix with #1838 (both edit `detect_within_timeframe` in overlapping ways, so they'd repeatedly re-conflict if merged separately). #1841 carries both fixes and both regression tests.

Issue #1753, opened 2026-05-04
Add unique alert count for MITRE ATT&CK tactics


YamatoSecurity:
@fukusuket Is it possible to put the count of unique alerts/detections to the right of the detected attack techniques?

<img width="852" height="523" alt="Image" src="https://github.com/user-attachments/assets/16c2d7d1-ce76-4745-822d-1ffff4f8b464" />

For example. `15. Impact (20)` instead of just `15. Impact`, etc...

If it is easy to also put the **total** number of alerts as well that would be also useful to know.

For example: `15. Impact (20 | 1000)`

That is, 20 unique alerts but a total of 1000 alerts.

fukusuket:
AI's struggling with the implementation, so I guess I'll code it by hand for a change :)

PR #1840, opened 2026-07-05
fix: scope MITRE report globals to one report + fix "Successed" typo

Shirofune-Security:
Two small cleanups surfaced while reviewing the MITRE tactic alert-count feature (#1753).

## 1. Scope the MITRE report accumulators to a single report

`_output_html_computer_by_mitre_attck` reads two process-global `lazy_static`s that are only ever appended to:
- `COMPUTER_MITRE_ATTCK_MAP` — per-computer `(tactic, unique, total)` counts
- `COMPUTER_MITRE_ATTCK_UNIQUE_KEYS` — the `computer|tactic|rule` keys added in #1753 to distinguish the **unique** count from the **total**

Neither is ever cleared. For the normal CLI path (one scan → one report → exit) this is harmless, but if a second report is generated in the same process — most plausibly across unit/integration tests — the retained keys make `insert()` return `false`, so the **unique** column is undercounted, and both structures grow unbounded.

Fix: clear both accumulators once the table has been emitted, so each report is self-contained. Also dropped an intermediate `Vec` when joining the tactic cells (`Itertools::join` directly on the iterator). **No behavior change for normal single-report runs.**

## 2. Fix the "Successed submodule update" typo

The message printed after a successful `update-rules` submodule pull read `Successed submodule update`; corrected to `Succeeded submodule update`.

## Testing
`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `RUST_TEST_THREADS=1 cargo test` (478 lib + 17 integration) all pass.

### DIFF
--- a/CHANGELOG-Japanese.md
+++ b/CHANGELOG-Japanese.md
@@ -1,5 +1,15 @@
 # 変更点
 
+## x.x.x [xxxx/xx/xx]
+
+**バグ修正:**
+
+- `update-rules` 実行後に表示される文法的に誤ったメッセージ「Successed submodule update」を「Succeeded submodule update」に修正した。 (#1840) (@YamatoSecurity)
+
+**その他:**
+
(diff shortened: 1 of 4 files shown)


YamatoSecurity (approved):
LGTM

Issue #1809, opened 2026-07-03
[bug] logon-summary: Target Domain is always "-" for RDS Gateway (EID 302) logons due to alias typo "RdsGtwUserName"


Shirofune-Security:
**Describe the bug**

In the `logon-summary` command, the `dst_domain` (Target Domain) extraction for RDS Gateway logon events (EID 302 in `Microsoft-Windows-TerminalServices-Gateway/Operational`) looks up the event key alias `"RdsGtwUserName"` (capital `N`), but the alias defined in `rules/config/eventkey_alias.txt` is `RdsGtwUsername` (lower-case `n`). Alias lookup is a case-sensitive `HashMap::get`, so the lookup misses and `get_event_value` falls back to `Event.EventData.RdsGtwUserName`, which never exists in these records (their fields live under `Event.UserData.EventInfo`). As a result, `dst_domain` is always `-` for RDS Gateway logons, even when the record's `Username` field contains `DOMAIN\user`. The neighboring `dst_user` arm uses the correct spelling `"RdsGtwUsername"` and works.

**Affected code**

The `dst_domain` arm with the typo (`RdsGtwUserName`):
https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/timeline/metrics.rs#L370-L378

```rust
                        RdsGtw => {
                            let user_with_domain = get_event_value_as_string(
                                "RdsGtwUserName",
                                &record.record,
                                &stored_static.eventkey_alias,
                            );
                            let domain = user_with_domain.rsplit_once('\\').map(|x| x.0);
                            CompactString::from(domain.unwrap_or("-"))
                        }
```

The `dst_user` arm just above uses the correct alias spelling:
https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/timeline/metrics.rs#L336-L347

```rust
                        RdsGtw => {
                            let user_with_domain = get_event_value_as_string(
                                "RdsGtwUsername",
                                &record.record,
                                &stored_static.eventkey_alias,
                            );
```

The only defined alias (note the lower-case `n`):
https://github.com/Yamato-Security/hayabusa-rules/blob/6dbc67d8446b719b40e3bcbf07b15dbb73dece1d/config/eventkey_alias.txt#L155

```
RdsGtwUsername,Event.UserData.EventInfo.Username
```

[...]

PR #1841, opened 2026-07-05
fix: enforce group-by equality and event order in temporal correlations

Shirofune-Security:
Combines two related fixes to Sigma temporal correlation evaluation (`detect_within_timeframe`) into one PR. **Supersedes #1838 and #1839** — they touch the same function in overlapping ways, so merging them separately would repeatedly re-conflict; this branch merges both fixes and both regression tests together.

## Fix 1 — group-by equality (was #1839, issue #1809)
Referenced-rule matches were correlated across **different group-by values**: rule A firing on `Host1` and rule B firing on `Host2` within the timeframe produced a false-positive temporal alert. Matches are now required to share the base result's `AggResult.key` (the `group-by` value).

## Fix 2 — event ordering (was #1838, issue #1810)
`temporal_ordered` enforced order only for the first referenced rule — a no-op `last_base = base` assignment left every later rule checked against the base event's window, so an out-of-order attack chain (`A→C→B` for a rule requiring `A→B→C`) still matched. Each referenced rule is now required to match **at or after the previous rule's matched event**, within the single timeframe window anchored at the base event.

## How they combine
The `temporal_ordered` branch applies both: it filters candidates to the base's group-by value, then picks the earliest one at/after the previous match within the window (`order_floor` advances per match; `window_end` stays anchored at the base). The unordered branch gets the group-by filter. The buggy `last_base` is removed.

## Testing
Both regression tests included:
- `test_detect_within_timeframe_enforces_group_by` — cross-group matches don't correlate; same-group do.
- `test_detect_within_timeframe_ordered_enforces_order` — in-order fires, out-of-order doesn't (ordered), still fires unordered, out-of-window doesn't.

`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `RUST_TEST_THREADS=1 cargo test` (480 lib + 17 integration) all pass.


YamatoSecurity:
@fukusuket sorry, i combined the PRs into one to make it easier to work with.

fukusuket (approved):
LGTM!!🚀

YamatoSecurity:
@fukusuket Thanks for checking!

Issue #1861, opened 2026-07-06
[refactor] Normalize DefaultMatcher pipe-modifiers into a MatchPlan; split matchers.rs


Shirofune-Security:
Port of the pro edition's refactor (hayabusa-pro#108). `src/detections/rule/matchers.rs` is ~4,700 lines; complexity concentrates in `DefaultMatcher::init`, a hand-enumerated, order- and index-sensitive table of pipe-modifier tuples (`pipes[0] == Base64 && pipes[1] == Contains`, `utf16|base64offset|contains`, `|contains|all|windash`, …) backed by the `PipeElement` enum. Adding one Sigma modifier touches several sites.

Two parts, mirroring pro:

1. **MatchPlan normalization** — fold the pipe list into a canonical `MatchPlan { wrap, encoding, utf16, cased, windash, all }` in one O(n) pass (`from_pipes` + `build_fast_match`) instead of matching hand-listed tuples; remove the now-subsumed `create_fast_match`. Output byte-identical on the sample-evtx corpus + all matcher tests. (hayabusa-pro#181)

2. **Module split** — break `matchers.rs` into a `matchers/` directory (`mod.rs` + `leaf_matchers.rs` + `default_matcher.rs` + `pipe_element.rs` + `tests.rs`). Pure code movement. (hayabusa-pro#182)

The per-modifier trait / category reorganization (`string/numeric/encoding/fieldref/cidr` with one impl per modifier) is a further follow-up beyond these two parts.

Shirofune-Security:
Both parts are up: **#1862** (part 1, MatchPlan normalization) and **#1863** (part 2, the `matchers/` module split, stacked on #1862). Each verified byte-identical to main on the full sample-evtx corpus (CSV + JSON) plus the matcher tests; part 2 additionally via a normalized code-diff. The remaining piece — the per-modifier trait so each modifier is one `impl` (the `string/numeric/encoding/fieldref/cidr` reorg) — is a behavioral dispatch rewrite, so it's left as a follow-up; keeping this open to track it.

Shirofune-Security:
Closing — the two parts scoped by this issue are implemented and merged:

- **Part 1 — MatchPlan normalization** → #1862 (folds the pipe list into a canonical `MatchPlan { wrap, encoding, utf16, cased, windash, all }` via `from_pipes` + `build_fast_match`; the subsumed `create_fast_match` is removed)
[...]

PR #1862, opened 2026-07-06
refactor: normalize DefaultMatcher pipe-modifiers into a MatchPlan

Shirofune-Security:
**Part 1 of #1861** (port of hayabusa-pro#181; the `matchers.rs` module split is part 2).

## What
`DefaultMatcher::init` built the fast matcher from a hand-enumerated, **order- and index-sensitive** table of pipe-count / pipe-tuple cases — `pipes[0] == Base64 && pipes[1] == Contains`, `|contains|all|windash`, `utf16|base64offset|contains`, `*|cased`, … — across nested `if n == 2 { … } else if n == 3 { … }` branches. This folds the pipe list into a canonical plan in one O(n) pass:

```rust
struct MatchPlan { wrap, encoding, utf16, cased, windash, all }
MatchPlan::from_pipes(&pipes).and_then(|p| p.build_fast_match(…))
```

The single-pipe helper `create_fast_match` is subsumed by the plan and removed.

## Behavior
- **Canonically-ordered rules (every real Sigma rule) are unchanged.** Detection output is **byte-identical to `main`** on the full sample-evtx corpus: `csv-timeline` (43,457,951 B) and `json-timeline -L` (47,987,522 B) both `cmp`-identical.
- **All 157 matcher tests + full lib suite (483) pass**; `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

### Order-independence (intended, tested)
The plan is order- and count-independent. A **non-canonical** modifier order the old table missed used to fall through to a wildcard regex that silently *ignored* the missed modifier; it is now normalized like its canonical form (pinned by new `test_reordered_*` / `test_utf16_base64_contains_*` tests):

| non-canonical key | old (fallback) | new (normalized) |
|---|---|---|
| `\|cased\|contains` | case-**in**sensitive | case-sensitive |
| `\|contains\|base64` | plaintext (ignored base64) | base64-encoded |
| `\|contains\|base64offset` | plaintext | base64offset |
| `\|base64\|utf16\|contains` | plaintext | utf16 + base64 |

Sigma fixes modifier order, so no real rule uses these — hence the byte-identical corpus result. This is strictly more correct than the old silent fallback, but it's a behavior change for malformed keys, so it's called out rather than claimed as "no change".


YamatoSecurity (approved):
LGTM

PR #1878, opened 2026-07-10
refactor: retire the COMPUTER_MITRE_ATTCK_MAP global and drop RUST_TEST_THREADS=1

Shirofune-Security:
## What

`COMPUTER_MITRE_ATTCK_MAP` (`DashMap`) and `COMPUTER_MITRE_ATTCK_UNIQUE_KEYS` (`DashSet`) were the last shared-mutable process globals — the per-computer MITRE ATT&CK tactic accumulators for the HTML report. They're written from the per-record **parallel** tasks (`create_log_record` in `execute_rule`) and read back when the report is rendered, then `.clear()`ed after emit only so a second report in the same process (e.g. across tests) wouldn't leak keys.

## How

- Move both onto `StoredStatic` as `Arc<DashMap<..>>` / `Arc<DashSet<..>>` fields, mirroring `pivot_keyword`/`error_log_stack`. The `Arc` lets the per-scan snapshot cloned into the parallel tasks share the same accumulators with the live `stored_static`.
- `create_log_record` writes `stored_static.computer_mitre_attck_*`; `_output_html_computer_by_mitre_attck` now takes `stored_static` and reads/clears the fields.
- A `ComputerMitreAttckMap` type alias keeps the field under clippy's `type_complexity` limit.
- **With `STORED_STATIC` already gone, no shared-mutable state remains to force serialized tests — `RUST_TEST_THREADS: 1` is dropped from both `rust.yml` and `coverage.yml`.**

## Verification

- `cargo fmt --all`, `cargo clippy --all-targets -- -D warnings` clean.
- Lib + bin suites pass **repeatedly under full parallelism** (no `RUST_TEST_THREADS=1`): 486 lib + 19 bin, multiple consecutive clean runs.
- **Behavior-preserving**: before vs. after binaries — `csv-timeline` output is byte-identical (52,388 lines), and the HTML report is byte-identical (incl. the **MITRE ATT&CK Tactics table** per-computer `unique | total` counts) apart from the volatile command-line / filename / elapsed-time lines.

This mirrors the same change in the hayabusa-pro edition. This PR's own CI (Windows build + coverage, tests now run in parallel) is the live check that dropping `RUST_TEST_THREADS=1` is safe.


Shirofune-Security:
Thanks for the review. Confirmed the parallel-safety concern:

- The 4 `detection.rs` tests that name `./test_emit_csv.csv` only set it as an unused `output` config path — they call `create_log_record` and assert on the in-memory `output_fields`, never `emit_csv`. `test_emit_csv_output` is the sole writer, so there's no collision.
- The other repo-root names that appear across multiple tests (`test-rule.yml`, `exclude_rules.txt`, `rule.yml`) are label/config-key strings passed to `create_rule` / the rule-exclude lists, never written to disk.
- `./logs/errorlog-<timestamp>.log` is only written from `main.rs` (production), never from a test, and is gated by `quiet_errors_flag`.

Together with CI passing under parallel tests (Windows build + coverage), dropping `RUST_TEST_THREADS: 1` is safe here.

I'll keep the test-I/O hardening out of this PR to keep the refactor focused, and open a separate follow-up that moves the file-output tests (`test_emit_csv_output`, the `overwrite_*` tests) onto per-test temp directories.

YamatoSecurity (approved):

Issue #1825, opened 2026-07-03
[bug] -G with a custom rules config dir (-c) ignores its geoip_field_mapping.yaml (lookup is missing the .yaml extension)


Shirofune-Security:
**Describe the bug**

When the GeoIP option (`-G`) is combined with a custom rules config directory (`-c/--rules-config`), the `geoip_field_mapping.yaml` file inside that directory is never found, because the lookup joins the config dir with the bare name `geoip_field_mapping` — without the `.yaml` extension:

https://github.com/Yamato-Security/hayabusa/blob/de07611599d47271076f34a4b0cd5a4848d6500f/src/detections/configs.rs#L323-L330

```rust
            let geo_ip_file_path = check_setting_path(config_path, "geoip_field_mapping", false)
                .unwrap_or_else(|| {
                    check_setting_path(
                        &CURRENT_EXE_PATH.to_path_buf(),
                        "rules/config/geoip_field_mapping.yaml",
                        true,
                    )
                    .unwrap()
                });
```

`check_setting_path` checks `base_path.join(path).exists()`, i.e. `<config dir>/geoip_field_mapping` — a file that does not exist (the real file is `geoip_field_mapping.yaml`). The all-in-one config branch misses for the same reason: `ONE_CONFIG_MAP` is keyed by full file names like `geoip_field_mapping.yaml`, and the basename passed in is `geoip_field_mapping`:

https://github.com/Yamato-Security/hayabusa/blob/de07611599d47271076f34a4b0cd5a4848d6500f/src/detections/utils.rs#L604-L615

As a result the lookup always falls through to the exe-relative default `rules/config/geoip_field_mapping.yaml`, silently ignoring the user's `-c` directory.

**Impact / example**

On a machine where the default `./rules` directory does not exist (or is out of date) and the user supplies a complete config directory explicitly:

```
hayabusa csv-timeline -f sample.jsonl -J -r ./my-rules -c /path/to/config -G /path/to/mmdb -o out.csv
```

even though `/path/to/config/geoip_field_mapping.yaml` exists, Hayabusa exits with:

```
[ERROR] Could not find the geoip_field_mapping.yaml config file. Please run update-rules.
```

[...]

Issue #1820, opened 2026-07-03
[bug] Splunk-JSON fallback timestamp parse discards the UTC offset, skewing First/Last Timestamp in log-metrics


Shirofune-Security:
**Describe the bug**

The fallback timestamp parser used to compute `First Timestamp` / `Last Timestamp` accepts Splunk-JSON style timestamps that carry a UTC offset (e.g. `2021-12-23T00:00:00.000+09:00`), but parses them with `NaiveDateTime::parse_from_str`. Per the chrono documentation, `NaiveDateTime::parse_from_str` validates the `%:z` field but discards it ("Offset is ignored for the purpose of parsing"). The resulting naive wall-clock time is then re-labeled as UTC via `DateTime::<Utc>::from_naive_utc_and_offset(..., Utc)`, so every non-UTC timestamp ends up shifted by its offset (9 hours for `+09:00`). The format string is exactly the Splunk `_time` export format that Hayabusa uses elsewhere for `is_splunk_json` input, so this affects Splunk-exported JSON analyzed with `-J`.

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/timeline/log_metrics.rs#L39-L57

```rust
let timestamp = match NaiveDateTime::parse_from_str(
    evttime.as_str(),
    "%Y-%m-%dT%H:%M:%S%.fZ",
) {
    Ok(without_timezone_datetime) => Some(
        DateTime::<Utc>::from_naive_utc_and_offset(without_timezone_datetime, Utc),
    ),
    Err(_) => {
        match NaiveDateTime::parse_from_str(
            evttime.as_str(),
            "%Y-%m-%dT%H:%M:%S%.3f%:z",
        ) {
            Ok(splunk_json_datetime) => {
                Some(DateTime::<Utc>::from_naive_utc_and_offset(
                    splunk_json_datetime,
                    Utc,
                ))
            }
```

The same pattern exists in `stats_time_cnt`, which produces the Results Summary `First Timestamp` / `Last Timestamp` and is fed Splunk `_time` values via its `Event.System.@timestamp` lookup:

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/timeline/metrics.rs#L136-L145

```rust
Err(_) => {
[...]

Issue #1817, opened 2026-07-03
[bug] eid-metrics: u16 underflow in Event column width calculation on terminals narrower than 55 columns


Shirofune-Security:
**Describe the bug**

In `tm_stats_dsp_msg()` (the `eid-metrics` command's Total/%/Channel/ID/Event table), the maximum width of the "Event" column is computed as `cmp::max(terminal_width - 55, 45)`. `terminal_width` is a `u16` taken from `terminal_size()`, so on an interactive terminal narrower than 55 columns the subtraction underflows before `cmp::max` can clamp the value to 45. In builds with overflow checks enabled (default dev/test builds) this panics with "attempt to subtract with overflow"; in the shipped release binaries (overflow checks off) it silently wraps around to a huge value (e.g. 65521 for a 40-column terminal), so the "Event" column is effectively unconstrained and the table overflows and wraps on exactly the narrow terminals the 45-character floor was meant to handle.

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/timeline/timelines.rs#L226-L237

```rust
        let terminal_width = match terminal_size() {
            Some((Width(w), _)) => w,
            None => 100,
        };

        let constraints = [
            LowerBoundary(Fixed(7)),  // Minimum number of characters for "Total"
            UpperBoundary(Fixed(9)),  // Maximum number of characters for "percent"
            UpperBoundary(Fixed(20)), // Maximum number of characters for "Channel"
            UpperBoundary(Fixed(12)), // Maximum number of characters for "ID"
            UpperBoundary(Fixed(cmp::max(terminal_width - 55, 45))), // Maximum number of characters for "Event"
        ];
```

`terminal_size::Width` wraps a `u16` and `comfy_table::Width::Fixed` also takes a `u16`, so `terminal_width - 55` is unchecked `u16` arithmetic. The `None => 100` fallback only applies when no attached tty can be found at all (`terminal_size()` falls back to stderr and stdin before returning `None`), so it does not protect narrow terminals.

**Impact / example**

Run `hayabusa.exe eid-metrics -f Security.evtx` in a terminal resized to fewer than 55 columns (e.g. 40):

[...]

PR #1897, opened 2026-07-14
fix: four log/config/metrics bugs (#1816, #1817, #1820, #1825)

### LINKED ISSUE #1816: [bug] "Cannot open file" error shows the literal placeholder `{path}` instead of the file path in `read_jsonl_to_value` / `read_json_to_value`
**Describe the bug**

In `src/detections/utils.rs`, both `read_jsonl_to_value` and `read_json_to_value` build their file-open error message from a plain string literal instead of `format!`, so the `{path}` placeholder is never interpolated. When a JSON input file cannot be opened (e.g. permission denied, or the file disappears between directory enumeration and open), the user literally sees `Cannot open file. [file:{path}]` with no indication of which file failed. Other error sites in the same file (e.g. `read_txt` at L109 and `read_csv` at L189) use `format!("Cannot open file. [file:{filename}]")` correctly, so this looks like a simple oversight.

**Affected code**

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/utils.rs#L119-L126

```rust
pub fn read_jsonl_to_value(path: &str) -> Result<Box<dyn Iterator<Item = Value>>, String> {
    let f = File::open(path);
    if f.is_err() {
        return Err("Cannot open file. [file:{path}]".to_string()); // literal, not format!
    }
    let reader = BufReader::new(f.unwrap());
    let mut peekable_lines = reader.lines().peekable();
    let first_line = peekable_lines.peek().unwrap();
```

https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/utils.rs#L145-L149

```rust
pub fn read_json_to_value(path: &str) -> Result<Box<dyn Iterator<Item = Value>>, String> {
    let f = fs::read_to_string(path);
    if f.is_err() {
        return Err("Cannot open file. [file:{path}]".to_string()); // literal, not format!
    }
```

For comparison, the same message is built correctly elsewhere in this file (https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/detections/utils.rs#L109 and #L189):

```rust
let errmsg = format!("Cannot open file. [file:{filename}]");
```

**Impact / example**

When scanning JSON input (`-J/--JSON-input`), `analysis_json_file` in `src/main.rs` calls `read_jsonl_to_value` and falls back to `read_json_to_value`; the latter's error is printed via `AlertMessage::alert` (https://github.com/Yamato-Security/hayabusa/blob/11ce47c6abf57ab67d1028b2dd3a199d0ff9de71/src/main.rs#L2543-L2557). So for an unreadable input file the user sees:

```
[ERROR] Cannot open file. [file:{path}]
```

instead of the actual file name — unhelpful when scanning a directory of many JSON files. The existing unit tests only assert `is_err()` on an invalid path, so the malformed message goes unnoticed.

Related notes in the same functions (separate from the message bug, happy to split into their own issue if preferred):
- `read_jsonl_to_value` calls `peekable_lines.peek().unwrap()` (L126), which panics on an empty (zero-line) file instead of returning `Err`.
- Inside the JSONL mapping closure, `serde_json::from_str(&line).unwrap()` (L136) panics if any line after the first is non-empty but invalid JSON; since the iterator is consumed lazily during the scan, a single malformed line aborts the whole run.

**Possible fix**

Change both sites to `format!("Cannot open file. [file:{path}]")`. Optionally, also handle the empty-file case with something like `let Some(first_line) = peekable_lines.peek() else { return Err(...) };` and replace the `unwrap()` in the mapping closure with a `filter_map` that skips (or reports) unparsable lines.

Shirofune-Security:
Batches four independent, well-scoped bug fixes, each with a regression test. Closes #1816, #1817, #1820, #1825.

| # | Bug | Fix |
|---|-----|-----|
| **#1816** | `read_jsonl_to_value`/`read_json_to_value` returned `"Cannot open file. [file:{path}]".to_string()` — a plain string, so the literal `{path}` was shown instead of the path | Use `format!` to interpolate `path` |
| **#1817** | eid-metrics capped the "Event" column at `cmp::max(terminal_width - 55, 45)`; on a terminal <55 cols the `u16` subtraction underflows → panic (debug) / wraps to ~65521 (release), defeating the 45-char floor | Extract `eid_metrics_event_col_width()` using `saturating_sub` |
| **#1820** | log-metrics parsed Splunk-JSON timestamps (`...+09:00`) with `NaiveDateTime`, discarding the offset and storing local time as UTC — skewing First/Last Timestamp | Extract `parse_log_metrics_timestamp()` that applies the offset via `DateTime::parse_from_str` + `with_timezone(&Utc)` |
| **#1825** | `-G` with a custom config dir (`-c`) ignored that dir's `geoip_field_mapping.yaml`: the lookup used the extensionless name `geoip_field_mapping`, always falling back to the bundled default | Look up `geoip_field_mapping.yaml` |

## Tests
Four new regression tests (each asserts the corrected behavior; the pre-fix code violates them):
- `test_read_json_open_error_includes_path` — errors contain the path, not `{path}`
- `test_eid_metrics_event_col_width` — 0/40/100/101/200-col widths, no underflow
- `test_parse_log_metrics_timestamp_applies_offset` — `+09:00`/`-05:00` applied
- `test_check_setting_path_geoip_yaml_extension` — `.yaml` name resolves in a custom dir, extensionless does not

## Verification
- **All 498 lib tests pass** (494 + 4 new); fmt clean.
- **Byte-identical** detection/metrics output on the sample-evtx corpus — `logm.csv` unchanged (the evtx `...Z` timestamp path is untouched), `eid.csv` content unchanged (#1817 only affects on-screen table width, not the `-o` file). The only harness diffs are the known HashMap-seed row-ordering in `eid.csv`/`logon-successful.csv`, verified sorted-identical.


Shirofune-Security:
Thanks — both addressed:

1. **CHANGELOG conflict:** rebased onto `main` and resolved the `CHANGELOG.md` / `CHANGELOG-Japanese.md` conflicts (the four `#1897` bug-fix entries now sit above the merged `#1896` entry). Merge state is clean again.

2. **`metrics.rs` still discarded Splunk offsets:** fixed. Rather than duplicate the parser, I extracted a shared `utils::parse_evtx_timestamp` (offset-aware: `DateTime::parse_from_str(...).with_timezone(&Utc)`) and routed all three call sites through it — `log_metrics.rs` plus `EventMetrics::stats_time_cnt` and `parse_evtx_datetime` in `metrics.rs` — so `eid-metrics`/`logon-summary` time ranges and the general first/last timestamps now apply the offset too, and the fix can't drift out of sync again. The parser unit test moved to `utils.rs` (covers `+09:00`/`-05:00`).

The evtx UTC (`...Z`) path is unchanged, so on the sample-evtx corpus every timeline/metrics output is byte-identical to `main` (eid/logon content-identical modulo the known HashMap-seed row order); only Splunk-JSON offset timestamps change. 498 lib tests pass, fmt clean.

YamatoSecurity (approved):

Issue #1906, opened 2026-07-21
Combine csv-timeline and json-timeline into a single dfir-timeline command


Shirofune-Security:
## Summary

Merge the `csv-timeline` and `json-timeline` commands into one **`dfir-timeline`** command whose output format is chosen with `-t, --output-type`. This removes the duplication between the two commands (they share almost all options and logic) and gives a single, consistent entry point for building a DFIR timeline.

## Proposed behavior

- **New command:** `dfir-timeline` (replaces `csv-timeline` and `json-timeline`).
- **`-t, --output-type <OUTPUT_FORMAT>`** selects the format:
  - default (unspecified) → **CSV**
  - `-t csv` → CSV (explicit)
  - `-t json` → JSON
  - `-t jsonl` → JSONL
  - **case-insensitive** (`-t JSONL`, `-t Json`, etc. all work)
- **`-t` shorthand moves to `--output-type`.** It is currently the shorthand for `--threads`; remove the `-t` shorthand from `--threads` (long `--threads` stays) and reuse `-t` for `--output-type`.
- **CSV-only options** — `-S, --tab-separator`, `-R, --remove-duplicate-data`, and `-M, --multiline` — only apply to CSV output:
  - group them under a separate **"CSV Output"** category in the help menu, and
  - **error** if any of them is set while the output format is not CSV (e.g. `-t json -M`).
- **Drop the `-L, --JSONL-output` flag** that `json-timeline` had — `-t jsonl` replaces it.
- **No multi-format output** — exactly one format per run (unlike suzaku, hayabusa does not support writing several formats at once).

## Example

```
hayabusa dfir-timeline -d ./logs                 # CSV (default)
hayabusa dfir-timeline -d ./logs -t json -o t.json
hayabusa dfir-timeline -d ./logs -t JSONL -o t.jsonl
hayabusa dfir-timeline -d ./logs -t json -M      # error: -M/--multiline only applies to CSV output
```

## Notes

- The vast majority of the two commands' `Action` match arms are identical pairs and collapse into a single `DfirTimeline` arm; only the output-writer selection and format flags become `output_type`-driven.
- Docs, the README command list, and the integration-test workflow need updating to the new command name.

A PR implementing this will follow.

PR #1907, opened 2026-07-21
feat: merge csv-timeline and json-timeline into a single dfir-timeline command

Shirofune-Security:
Closes #1906.

## What

Combines `csv-timeline` and `json-timeline` into one **`dfir-timeline`** command; the output format is chosen with **`-t, --output-type`**.

```
hayabusa dfir-timeline -d ./logs                    # CSV (default)
hayabusa dfir-timeline -d ./logs -t json  -o t.json
hayabusa dfir-timeline -d ./logs -t JSONL -o t.jsonl   # case-insensitive
```

- `-t, --output-type <OUTPUT_FORMAT>`: `csv` (default), `json`, `jsonl` — **case-insensitive** (`-t JSONL`, `-t Json`).
- **`-t` moved from `--threads` to `--output-type`**; `--threads` keeps its long form (its `-t` shorthand is removed).
- **CSV-only options** (`-M, --multiline`, `-S, --tab-separator`, `-R, --remove-duplicate-data`) are grouped under a new **"CSV Output"** help category and now **error** when combined with a non-CSV `-t`:
  ```
  $ hayabusa dfir-timeline -t json -M -o t.json
  [ERROR] The following option(s) only apply to CSV output and cannot be used with the selected -t, --output-type: -M, --multiline
  ```
- **`-L, --JSONL-output` is removed** — use `-t jsonl`.
- No multi-format output (one format per run), per the requested scope.

## Implementation notes

- The two `Action` variants collapse into one `DfirTimeline(DfirTimelineOption)` — the vast majority of the paired `match` arms were byte-identical and merged automatically. Output-writer selection (`init_writer`) and the json/jsonl format flags (`emit_csv_inner`) are now driven by `output_type`, and the structured `#AllFieldInfo` build (previously gated on json-timeline) now triggers for json/jsonl but **not** csv — so CSV output is unaffected.
- Updated the `integration-test` and `timeline-diff` workflows (including old `-t <N>` **thread counts** → `--threads <N>`, which would otherwise now parse as an output format) and the `doc/` command examples.

## Verification

- `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the full test suite (**501 + 20**) all pass.
- Functional: all of `-t csv|json|jsonl` (and uppercase) produce the right format; `-t csv` == default; the CSV-only-options validation fires; `-M` with CSV is fine; `csv-timeline`/`json-timeline` are now rejected as unknown subcommands.
[...]

PR #1908, opened 2026-07-22
docs: update website command reference for the dfir-timeline merge + make every subcommand's help menu consistent

Shirofune-Security:
Documentation follow-up to the `csv-timeline` + `json-timeline` -> `dfir-timeline` merge, which is already on `main`. Companion to #1906.

Rebased onto current `main` and resynced with the CLI as of `hayabusa v4.0.0`.

## What

- **`website/docs/commands/dfir-timeline.md` + all 14 localized pages**: merged the separate `csv-timeline` and `json-timeline` sections into one `dfir-timeline` section — embedded the current `dfir-timeline --help` output, documented **`-t, --output-type`** (`csv` default / `json` / `jsonl`, case-insensitive), kept the "when to use CSV vs JSON vs JSONL" guidance, added `-t json` and `-t jsonl` examples, and removed the dropped `-L, --JSONL-output`. The help block now also reflects #1909, which moved `-R, --remove-duplicate-data` out of `CSV Output:` into `Output:` — so the prose about CSV-only options lists just `-M` and `-S`.
- **The rest of the site (all languages)**: command examples/names updated — `csv-timeline` → `dfir-timeline`, `json-timeline` → `dfir-timeline -t json`, `json-timeline … -L` → `dfir-timeline … -t jsonl`, and old `-t <N>` **thread counts** → `--threads <N>` (they would otherwise now parse as an output format).

## Collateral damage from the rename, undone

The bulk rename caught three things it should not have:

- **The changelog is a historical record.** Entries describing what `csv-timeline` and `json-timeline` did at the time have been restored, including several that had been rewritten to read "`dfir-timeline` and `dfir-timeline`".
- **`split-csv-timeline` belongs to [Takajo](https://github.com/Yamato-Security/takajo), not Hayabusa.** It had been renamed to `split-dfir-timeline`, which also broke the anchor on the link to Takajo's README.
- **The command index listed `dfir-timeline` twice** — once "in CSV format", once "in JSON/JSONL format". Collapsed into a single entry covering all three formats.

## Examples that no longer ran as written

- `dfir-timeline -t json … -J -o results.jsonl` does not produce JSONL any more: `-J` is now `--json-input`. Changed to `-t jsonl`. The same sentence claimed "in the current 2.x version of Hayabusa", which is dropped.
- The `expand-list` and `extract-base64` examples both invoked `eid-metrics`, and the `extract-base64` CSV example passed `-r`, which that command does not accept.
- The `log-metrics` examples described their output as "Event ID metrics" (that is `eid-metrics`) and wrote to `eid-metrics.csv`.

## Consistent help menus (`src/detections/configs.rs`)

This is the one non-docs change in the PR, and the docs above are generated from it.

Five subcommands rendered `--help` differently from the other ten. `list-contributors` and `list-profiles` had no `help_template`, so they printed neither the version banner nor the author line and put the usage on a single line. `level-tuning`, `set-default-profile` and `update-rules` had the banner but interpolated clap's `{usage}`, which renders the running binary's name — `hayabusa` on Linux and macOS — where the other ten hardcode `hayabusa.exe`.

All fifteen now open the same way:

```
Hayabusa v4.0.0 - Black Hat USA Arsenal 2026 Release
Yamato Security (https://github.com/Yamato-Security/hayabusa - @SecurityYamato)

Usage:
  hayabusa.exe <command> [OPTIONS]
```

`expand-list` also advertised `<INPUT>` while accepting no input option — it has no `-d`/`-f`/`-l`, only `-r, --rules` — so that is dropped from its usage line.

The 210 embedded help blocks across all 15 languages are regenerated to match. Their old usage lines came from three different eras and were inconsistent in exactly the same way the CLI was.

## Verification

Checked against a release build of current `main`:

- for all 15 subcommands, `-h`, `--help` and `help <command>` agree;
- every English help block matches its command's `--help` **byte for byte** once the version banner is stripped (210 blocks checked, 14 of the 15 subcommands are documented; `list-contributors` has no section on the site);
- every localized block matches **structurally** — same usage line, sections, option specs and order, with only the descriptions translated;
- all **750** example invocations across the docs (1620 option tokens) use a real subcommand with options that subcommand accepts;
- no `csv-timeline`/`json-timeline` references and no old capitalized long-names (`--GeoIP`, `--UTC`, …) remain under `website/docs/`.

## Note

The localized prose was produced with machine translation matched to each page's existing style; **native-speaker review is welcome** on the merged intro paragraphs. The code blocks, flag names, and command names are identical across all languages.


Shirofune-Security:
Note: the code PR this documents is now **#1909** (#1907 was moved to a branch on this repo for full CI access and closed).

YamatoSecurity (approved):

PR #1909, opened 2026-07-22
feat: merge csv-timeline and json-timeline into a single dfir-timeline command

### LINKED ISSUE #1906: Combine csv-timeline and json-timeline into a single dfir-timeline command

Shirofune-Security:
Closes #1906. Same changes as #1907, re-opened from a branch on this repo (instead of a fork) so the full CI/secrets are available; #1907 will be closed in favor of this one.

## What

Combines `csv-timeline` and `json-timeline` into one **`dfir-timeline`** command; the output format is chosen with **`-t, --output-type`**.

```
hayabusa dfir-timeline -d ./logs                    # CSV (default)
hayabusa dfir-timeline -d ./logs -t json  -o t.json
hayabusa dfir-timeline -d ./logs -t JSONL -o t.jsonl   # case-insensitive
```

- `-t, --output-type <OUTPUT_FORMAT>`: `csv` (default), `json`, `jsonl` — **case-insensitive**.
- **`-t` moved from `--threads` to `--output-type`**; `--threads` keeps its long form.
- **CSV-only options** (`-M, --multiline`, `-S, --tab-separator`, `-R, --remove-duplicate-data`) are grouped under a new **"CSV Output"** help category and **error** when combined with a non-CSV `-t`.
- **`-L, --JSONL-output` removed** — use `-t jsonl`. One format per run.

## Implementation notes

The two `Action` variants collapse into one `DfirTimeline` (the paired match arms were identical); output-writer selection and the json/jsonl format flags are driven by `output_type`, and the structured `#AllFieldInfo` build now triggers for json/jsonl but **not** csv (so CSV output is unaffected). Updated the `integration-test`/`timeline-diff` workflows (including old `-t <N>` thread counts → `--threads <N>`) and `doc/` examples.

## Verification

- `cargo build`, `clippy --all-targets -D warnings`, `cargo fmt --check`, full test suite (**501 + 20**) all pass; Windows CI green (on #1907).
- **Byte-identical regression** over the 599-file `hayabusa-sample-evtx` corpus: old `csv-timeline` == new `dfir-timeline`, old `json-timeline` == `dfir-timeline -t json`, old `json-timeline -L` == `dfir-timeline -t jsonl`.
- Functional: all `-t csv|json|jsonl` (and uppercase); CSV-only-options validation fires; `csv-timeline`/`json-timeline` are now unknown subcommands.

The website command-reference docs (translated) are updated in the follow-up PR #1908.


Shirofune-Security:
I found one workflow bug to fix:

`.github/workflows/timeline-diff.yml` switches back to `main` at line 56, but the commands in the `Run on main branch` and `Run on main branch(encoded_rules)` steps were also changed to `dfir-timeline` (lines 58-59 and 70-71). The current `main` branch still has `csv-timeline` / `json-timeline`, not `dfir-timeline`, so this workflow will fail whenever it reaches the baseline side of the comparison.

Please keep the `main`-branch commands on the old CLI, e.g. `csv-timeline --threads 1 ...` and `json-timeline --threads 1 ... -L`, while using `dfir-timeline` only for the PR/dev side. That still compares old output vs new output and keeps the workflow runnable before the PR is merged.

Local checks passed after initializing the rules submodule: `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --locked` (501 lib tests + 20 main tests).

YamatoSecurity (approved):

src/main.rs
```diff
@@ -449,6 +449,30 @@ impl App {
...
+            && !matches!(opt.output_type, OutputType::Csv)
+        {
+            let mut csv_only = vec![];
+            if opt.multiline {
+                csv_only.push("-M, --multiline");
+            }
+            if opt.tab_separator {
+                csv_only.push("-S, --tab-separator");
+            }
+            if opt.output_options.remove_duplicate_data {
```
Shirofune-Security: `-R, --remove-duplicate-data` should not be included in `csv_only`. This is a compatibility regression: v3.10 supports `json-timeline -R -s`, the changelog explicitly documents duplicate suppression for JSON, and the merged JSON/JSONL renderer still passes `remove_duplicate_data` into `output_json_str` (`src/results/csv.rs`, lines 91-96 and 166-174). With this validation, the equivalent v4 command produces no timeline. Please keep `-R` available for CSV, JSON, and JSONL, leaving only `-M` and `-S` in this check. If removing JSON duplicate suppression is intentional, document it as a breaking removal and delete the now-unreachable JSON behavior/tests instead.

src/main.rs
```diff
@@ -449,6 +449,30 @@ impl App {
...
+                csv_only.push("-R, --remove-duplicate-data");
+            }
+            if !csv_only.is_empty() {
+                AlertMessage::alert(&format!(
+                    "The following option(s) only apply to CSV output and cannot be used with the selected -t, --output-type: {}",
+                    csv_only.join(", ")
+                ))
+                .ok();
+                return;
```
Shirofune-Security, line 490: This invalid invocation exits successfully. `App::exec` returns `()`, so this plain `return` reaches the end of `main()` and the process exits with status 0. I reproduced `dfir-timeline -t json -M ...` printing `[ERROR]` while `$?` was `0`; no output was created. That will make scripts and CI treat a rejected command as successful. Please enforce the format conflicts during clap parsing, or return a `Result`/`ExitCode` and propagate a nonzero status here.

doc/AnalysisWithJQ-English.md
```diff
@@ -142,7 +142,7 @@ The JSON format is ideal if you are also going to do some manual checking.
 
 # Creating JSON Results Files
 
-In the current 2.x version of Hayabusa, you can save the results in JSON with `hayabusa json-timeline -d <directory> -o results.json` or `hayabusa json-timeline -d <directory> -J -o results.jsonl` for JSONL format.
+In the current 2.x version of Hayabusa, you can save the results in JSON with `hayabusa dfir-timeline -t json -d <directory> -o results.json` or `hayabusa dfir-timeline -t json -d <directory> -J -o results.jsonl` for JSONL format.
```
Shirofune-Security: The JSONL example selects the wrong output and repurposes `-J`. `-t json` selects regular JSON, while `-J` means JSON **input**; with an EVTX directory it changes the input filter to `.json`/`.jsonl` files instead of enabling JSONL output. The command should be `hayabusa dfir-timeline -t jsonl -d <directory> -o results.jsonl` (without `-J`). Please make the same correction in `AnalysisWithJQ-Japanese.md:150`, and update the stale `current 2.x` wording for v4.

doc/TimesketchImport/TimesketchImport-English.md
```diff
@@ -99,14 +99,14 @@ After you click `+ ADD TIMELINE`, you will see a dialog box asking you to upload
...
 ```
 
-> Note: It is necessary to choose a `timesketch*` profile and specify the timestamp as `--ISO-8601` for UTC or `--RFC-3339` for local time. You may add other Hayabusa options if you desire, however, do not add the `-M, --multiline` option as the newline characters will corrupt the import.
+> Note: It is necessary to choose a `timesketch*` profile and specify the timestamp as `--iso-8601` for UTC or `--rfc-3339` for local time. You may add other Hayabusa options if you desire, however, do not add the `-M, --multiline` option as the newline characters will corrupt the import.
 
 At the "Select file to upload" dialog box, name your timeline something like `hayabusa`, choose the `Comma (,)` CSV delimiter and click `SUBMIT`.
 
-> If your CSV file is too big to upload, you can split the file into multiple CSV files with Takajo's [split-csv-timeline](https://github.com/Yamato-Security/takajo?tab=readme-ov-file#split-csv-timeline-command) command.
+> If your CSV file is too big to upload, you can split the file into multiple CSV files with Takajo's [split-dfir-timeline](https://github.com/Yamato-Security/takajo?tab=readme-ov-file#split-dfir-timeline-command) command.
```
Shirofune-Security: This rename crosses a project boundary: Takajo has not renamed this command. Its current CLI still defines `split-csv-timeline`, and neither `split-dfir-timeline` nor the new anchor exists, so the guide now sends users to a broken link and nonexistent command. Please restore `split-csv-timeline` and `#split-csv-timeline-command` here, and restore the command name in the Japanese guide as well. Takajo definition: https://github.com/Yamato-Security/takajo/blob/ea8c6e43dc9158f4db06df4da7cfffc8589fbcfe/src/takajo.nim#L273

src/main.rs
```diff
@@ -3663,7 +3682,7 @@ mod tests {
         // Create an empty file first.
         let mut app = App::new(None);
         File::create(&out_overwrite_json_exit_json).ok();
-        let action = Action::JsonTimeline(JSONOutputOption {
+        let action = Action::DfirTimeline(DfirTimelineOption {
```
Shirofune-Security, line 3710: These migrated JSON overwrite tests now construct `DfirTimelineOption` without setting `output_type`, so `OutputType::default()` makes them CSV tests despite their names and `.json` paths. The same issue occurs in `test_overwrite_json` below. Please set `output_type: OutputType::Json` in both fixtures (and ideally enable the currently TODO-disabled assertions) so writer selection and JSON overwrite behavior remain covered.

Shirofune-Security (commented):
Additional review findings that cannot be attached cleanly to a changed line:

1. The previously reported baseline failure is still unresolved: after `git checkout main`, `.github/workflows/timeline-diff.yml` invokes `dfir-timeline`, which does not exist on the base revision. Please retain `csv-timeline` and `json-timeline -L` for the main-side baseline. See the existing report: https://github.com/Yamato-Security/hayabusa/pull/1909#issuecomment-5041394584

2. The encoded JSONL comparison at lines 98-100 repeats `diff main.jsonl dev.jsonl`, so `main-encoded.jsonl` and `dev-encoded.jsonl` are never compared. Please change it to `diff main-encoded.jsonl dev-encoded.jsonl`.

3. The bundled rules revision still has `rule-parse-error-check.yaml` and `update-sigmarule.yaml` workflows that dynamically download the latest Hayabusa release and invoke the removed `csv-timeline` command. Both workflows will fail as soon as v4 becomes the latest tag. Please update those workflows upstream to `dfir-timeline` and then bump the rules gitlink before the v4 release.

Local validation at this head: `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings`, and `cargo test --locked` all pass (501 library tests and 20 binary tests).

src/main.rs
Shirofune-Security: Fixed in `fed98258` — `-R` is out of `csv_only`, only `-M`/`-S` remain, and it is back under the "Output" help heading with the "(CSV output only, ...)" wording dropped.

Your point about the renderer still passing `remove_duplicate_data` into `output_json_str` is what settled it: the gate was the only thing stopping it, so the JSON path would have been live but unreachable.

One thing had to land first, though. Restoring `-R` for JSON as-is would have restored a **broken** feature. `output_json_str` decided a duplicate on `(profile string matches) || (resolved values match)`, and under JSON output `parse_message` never substitutes `%alias%` — so a `Details` profile holds the rule's *template*, identical for every event matching that rule, and `ExtraFieldInfo` is the constant `"-"`. Any two consecutive detections of one rule compared equal and the later one's real value was overwritten:

```
{ ... "Computer":"comp1","Details":{"Cmd":"whoami.exe"} }
{ ... "Computer":"comp2","Details":"DUP" }   <- actual value: netcat-backdoor.exe
```

So the same commit also fixes the comparison to read the substituted values from `details_convert_map`. Verified on the DeepBlueCLI corpus: `-t jsonl -R -s` gives 18,419 records (same as the baseline), 15,047 suppressions, and every one checked against its predecessor — **0 replaced a distinct value**. Genuine duplicates still collapse.

This is a port of Yamato-Security/hayabusa-pro#299 + #300, which hit exactly this sequence.

src/main.rs
Shirofune-Security: Fixed in `fed98258`, and it was wider than the one site — **nine** abort paths had it, all exiting 0:

- the CSV-only option check (this one)
- a rule-config error
- redundant `-A`/`-a`
- rules directory does not exist
- no .evtx files found
- `-f` file does not exist
- `-f` given a non-.evtx file
- no rules loaded

Same shape in each: report the problem, `return`, and because `App::exec` returns `()` and `main` then ends normally, the status stays 0. I confirmed there is no `process::exit` anywhere on those paths.

I did not use `process::exit` at the sites, and not clap-level validation either. Several unit tests deliberately drive `exec` into these paths and assert the run produced nothing (`test_same_file_output_*_exit` and friends) — `process::exit` there would kill the test process, and the checks are cross-field (`-t` versus `-M`/`-S`), which clap cannot express cleanly. Instead `App` carries a `failed` flag that the abort paths set and `main` turns into `exit(1)` after the runtime shuts down, so `exec` still returns normally under test.

Verified:

```
$ hayabusa dfir-timeline -t json -M ... ; echo $?
[ERROR] The following option(s) only apply to CSV output ...
1
$ hayabusa dfir-timeline -t csv ... ; echo $?
0
```

doc/AnalysisWithJQ-English.md
Shirofune-Security: Fixed in `fed98258`. Confirmed `-J` is `--json-input` ("Scan JSON formatted logs instead of .evtx (.json or .jsonl)"), so on an EVTX directory it changed the *input* filter rather than producing JSONL — the command as written could not have worked.

Both languages now read `hayabusa dfir-timeline -t jsonl -d <directory> -o results.jsonl`, with no `-J`, and the stale "current 2.x version" / 「現在バージョン2.x」 wording is dropped (English:145, Japanese:150).

doc/TimesketchImport/TimesketchImport-English.md
Shirofune-Security: Fixed in `fed98258` — restored to `split-csv-timeline` and `#split-csv-timeline-command` in both guides.

Verified against a local Takajo checkout rather than taking it on trust: `src/takajo.nim:273` defines `cmdName = "split-csv-timeline"`, matching the line you cited, and `split-dfir-timeline` appears nowhere in the CLI. So the link and the command name were both dead.

Worth noting the general point for the rest of this PR: renaming hayabusa's own subcommand does not rename anything in Takajo, and the docs are the only place the two projects meet.

src/main.rs
Shirofune-Security: Fixed in `fed98258` — both fixtures now set `output_type: OutputType::Json`.

The cause was the trailing `..Default::default()`: with `output_type` unset it filled in `OutputType::default()`, i.e. CSV. So two tests named `..._json`, writing `.json` paths, were exercising the CSV writer and the JSON writer-selection path had no coverage at all.

I left the `// TODO add check` assertions alone. They are commented out on the CSV originals these were migrated from, so enabling them is a pre-existing gap rather than something this PR introduced, and `MESSAGES.len()` is global state that would make the assertion order-dependent under parallel test runs. Happy to do it as a follow-up if you want the coverage.

Issue #1295, opened 2024-03-05
New `sort-csv` command


YamatoSecurity:
Since `--low-memory-mode` can not sort or remove duplicate entries it would be nice to have a command that can do this in post-processing. `sort-csv`: `sort and remove duplicate detections`

@hitenkoku Since you did the `-X, --remove-duplicate-detections`, could I ask you to do this one?
We need to keep the CSV header at the top and sort by timestamp, then remove duplicate entries but ignore the EvtxFile column if there is one because sometimes different `.evtx` files will have the same records. (backup files, etc...) I think this is the same logic as `-X, --remove-duplicate-detections`, is that correct?

I think we just need `-i, --input <FILE>`, `-o, --output <FILE>` and `-C, --clobber  Overwrite files when saving` options.

YamatoSecurity:
@hitenkoku In order to sort without using a lot of memory, it might be good to import the CSV data into a temporary sqlite database, sort the sqlite database and then export the results out back to CSV. What do you think?

YamatoSecurity:
So that an investigator can collect many CSV files from endpoints with velociraptor we should also support directory input. 
`-f, --file <FILE>  Input file` and `-d, --directory <DIRECTORY>   Input directory`

PR #1919, opened 2026-08-18
feat: add sort-csv command to sort and de-duplicate a Hayabusa CSV (#1295)

ChrisJr404:
Saw #1295 was still open, took a stab at it.

Adds a `sort-csv` command that post-processes a Hayabusa CSV. It keeps the header, sorts rows by the `Timestamp` column, and drops duplicate detections comparing every column except `EvtxFile`, so the same event pulled from overlapping or backup `.evtx` files collapses to one row. Same dedup rule as `-X, --remove-duplicate-detections`, just done after the fact on the CSV instead of during the scan (which is the gap `--low-memory-mode` leaves).

Options match what you laid out on the issue:

- `-f, --file` for a single CSV
- `-d, --directory` to sort a whole folder of CSVs together (e.g. a batch pulled off endpoints with Velociraptor)
- `-o, --output` to write to a file, otherwise it prints to screen
- `-C, --clobber` to overwrite

Timestamp parsing handles the formats `format_time` can emit (default, `--rfc-3339`, `--iso-8601`, US/European, `--rfc-2822`), normalizing to UTC for the compare. If a column carries a format I didn't cover it falls back to a lexical sort on the raw string so it still produces stable output. The sort is a total order (timestamp, then the whole row) so runs are deterministic.

On the temp-sqlite idea from the issue thread: left it out for v1 and did a straightforward in-memory sort. Wanted to get the behavior right first. The sqlite path for low-memory sorting can slot in behind the same command later if you want it.

Wired it in as a new `Action::SortCsv` handled up front in `main.rs`, since it needs none of the rule/profile/evtx setup the other subcommands do. New module lives at `src/timeline/sort_csv.rs` with a test covering sort order, header preservation, and the EvtxFile-ignoring dedup.

Tested against sample CSVs (single file with `-f`, and a folder with `-d`) and confirmed the sort plus cross-file dedup. `cargo test` for the module and `cargo fmt --check` are clean.

Refs #1295


YamatoSecurity:
@ChrisJr404 Thanks for this, could you check and resolve copilot comments?
@fukusuket Could you review  this as well? 🙏

ChrisJr404:
Thanks for the review. Addressed Copilot's point: the sort key now comes from `timestamp()` seconds plus subsecond nanos as an `i128` instead of `timestamp_nanos_opt()`, so timestamps outside the ~1677-2262 nanosecond range no longer return `None` and flip the whole file to the lexical fallback. Added a regression test that sorts a pre-1677 date ahead of a modern one.

fukusuket (commented):
@ChrisJr404 Thank you for the PR!  
I would expect the output of the `sort-csv` command to match the output of `dfir-timeline -s -X` without any differences, but there are differences as shown below. Do you know why this happens?
```
% ./hayabusa-4.0.0-mac-aarch64 dfir-timeline -d ../data/windows/hayabusa-sample-evtx-main -w -q -C -s -X -o timeline.csv
% ./hayabusa-1295 sort-csv -f timeline.csv -o sorted.csv
% diff timeline.csv sorted.csv | wc -l 
   12763
```

ChrisJr404:
Thanks for testing this. A large diff there is expected rather than a bug, and it comes from two things the `sort-csv` command does that `dfir-timeline -s -X` does not:

1. De-duplication: `sort-csv` drops duplicate rows, so if the timeline contains repeated events the output will simply have fewer lines. That alone shows up as a big `diff`.
2. Sort stability/key: `sort-csv` sorts the CSV rows directly, which can order rows that share the same timestamp differently from the order `dfir-timeline` produced internally, even though both are chronologically sorted.

Quick way to separate the two: compare `wc -l` of the two files. If `sorted.csv` has fewer lines, the difference is de-duplication; if the counts match but rows are reordered, it is the tie-break ordering.

If the goal is for `sort-csv` to be a faithful re-sort of an existing timeline, I can (a) make de-duplication opt-in via a flag rather than always-on, and (b) align the sort key and tie-break with `dfir-timeline` so equal-timestamp rows keep the same relative order. Let me know which behavior you'd prefer and I'll adjust.

fukusuket:
@ChrisJr404 
Since I ran it with the `-X` option, there was no difference in the line count. I compared only the timestamp column using awk '{print $1}' and confirmed they match. Thank you!

It looks like we can improve performance as noted below. 
Could you try applying these changes?


Profiled this on large timelines. Peak RSS is ~4.6x the input size, linear, unbounded (`-d` = 4.6x the whole directory).
Release build, 11-column CSV (`verbose` profile columns), `/usr/bin/time -l`:

| rows | file | wall | peak RSS |
|---|---|---|---|
| 500k | 167 MB | 1.39s | 778 MB |
| 2M | 674 MB | 6.78s | 3.10 GB |

At 500k: `rows: Vec<Vec<String>>` ≈ 460 MB, `seen: HashSet<Vec<String>>` ≈ 315 MB (40% of peak).

**1. `seen` holds a second full copy of every unique row.** `dedup_key` clones every cell. The key can borrow, and since `Timestamp` is part of the key, duplicates always share a timestamp and are adjacent after the sort — so the set only needs one timestamp group at a time. Same approach as `get_duplicate_indices` in `src/results/mod.rs:401`.

```rust
let mut seen: HashSet<Vec<&str>> = HashSet::new();
let mut prev_ts: Option<&str> = None;
for &idx in &order {
    let row = &rows[idx];
    let ts = row.get(timestamp_idx).map(String::as_str);
    if ts != prev_ts { seen.clear(); prev_ts = ts; }
    let key: Vec<&str> = match evtx_idx {
        Some(i) => row.iter().enumerate().filter(|(j, _)| *j != i).map(|(_, c)| c.as_str()).collect(),
        None => row.iter().map(String::as_str).collect(),
    };
    if seen.insert(key) { output_rows.push(row); }
}
```

**2. Keep `csv::StringRecord` instead of converting to `Vec<Vec<String>>`.** `StringRecord` is 2 allocations per row; the conversion in `read_csv` makes it 12 for an 11-column file. `records()` already returns it, so this is mostly deleting `.iter().map(str::to_string).collect()`. Comparison becomes `a.iter().cmp(b.iter())`.

**3. Simplify the sort.** Three things in one place: `all_parsed` is re-evaluated on every comparison; `parsed` is `Vec<Option<i128>>` where `Option<i128>` has no niche, so it's 32 bytes/row for 16 bytes of data; and the full-row tie-break `rows[a].cmp(&rows[b])` runs a deep comparison O(n log n) times, which isn't needed because the input order is already deterministic (`collect_input_files` sorts by path, rows follow file order) — a stable sort keyed on the timestamp alone preserves determinism.

```rust
const UNPARSED: i128 = i128::MIN;
let mut all_parsed = true;
let parsed: Vec<i128> = rows.iter()
    .map(|row| match row.get(timestamp_idx).and_then(|t| parse_timestamp(t)) {
        Some(v) => v,
        None => { all_parsed = false; UNPARSED }
    })
    .collect();
let mut order: Vec<usize> = (0..rows.len()).collect();
if all_parsed { order.sort_by_key(|&i| parsed[i]); }
else { order.sort_by_key(|&i| rows[i].get(timestamp_idx)); }
```

Dropping the tie-break means the surviving row of a duplicate group is picked by input order (i.e. path order) rather than by row content. Row count is unchanged, and duplicates differ only in `EvtxFile`, so this only affects which `EvtxFile` value is kept. Worth stating in a comment and a test so it's an explicit rule rather than an emergent property of the comparator.

## Measured

Applied to an extracted copy of the pipeline; the "before" is byte-identical in output to `hayabusa sort-csv` and within a few percent of its RSS and wall time (same mimalloc allocator).

Contribution of each change, 500k rows:

| | peak RSS | wall |
|---|---|---|
| current | 739 MB | 1.30s |
| (1) only | 372 MB (−50%) | 0.95s |
| (2) only | 661 MB (−11%) | 1.16s |
| (3) only | 742 MB (±0%) | 1.13s |
| all three | 286 MB (−61%) | 0.71s (1.8x) |

(2) looks weak alone only because the dedup copy dominates until (1) lands; together they compound. (3) is speed-only.

| input | current | all three |
|---|---|---|
| 500k / 167 MB | 739 MB / 1.30s | 286 MB / 0.71s |
| 2M / 674 MB | 2.89 GB / 6.18s | 1.11 GB / 3.19s |

4.6x → ~1.7x of input size. Memory stays linear in input either way — external merge sort would be the real fix for timelines larger than RAM, but that's out of scope here.

YamatoSecurity (approved):
LGTM! @ChrisJr404 Thanks for this!
@fukusuket Can you do the performance enhancement in another PR?

PR #1924, opened 2026-09-05
perf(sort-csv): borrow dedup keys, keep StringRecord, and simplify the sort

### LINKED ISSUE #1923: Improve `sort-csv` performance
Improve `sort-csv` performance as described here: https://github.com/Yamato-Security/hayabusa/pull/1919#issuecomment-5473368973

@fukusuket Could you do this one whenever you get time?

fukusuket:
Closed #1923 

Follow-up to #1295. `sort-csv` held peak RSS at about 3.9x the input size, linear and unbounded (`-d` means 3.9x the whole directory). This brings it down to about 1.7x and shortens the wall time. All changes are inside `src/timeline/sort_csv.rs`.

## Changes

1. **The dedup set borrows instead of cloning.** `dedup_key` copied every cell of every row into `seen`, a second full copy of the file. The key contains `Timestamp`, so duplicates always share a timestamp and are adjacent after the sort: `seen` can hold `&str` and only needs one run of equal keys at a time. Same approach as `get_duplicate_indices` in `src/results/mod.rs`. The sort and the dedup live in one function driven by a single `key_of` closure, because the reset is only safe while both use the same key.
2. **Rows stay as `csv::StringRecord`** (2 allocations per row) instead of being converted to `Vec<String>` (12 for an 11-column file).
3. **Simplified sort.** `all_parsed` is no longer re-evaluated per comparison, `parsed` is `Vec<i128>` rather than `Vec<Option<i128>>` (no niche, 32 bytes for 16 bytes of data), and the full-row tie-break is gone.

## Behavioral change

Dropping the tie-break makes rows that share a timestamp come out in input order (path order) rather than lexical order of their contents. Row count and the set of surviving detections are unchanged; duplicates differ only in `EvtxFile`, so the visible difference is which `EvtxFile` survives. Pinned by `rows_sharing_a_timestamp_keep_their_input_order`.

## Benchmarks

Release build, macOS, `/usr/bin/time -l`, best of 5. 11-column CSV, 5% duplicates differing only in `EvtxFile`, timestamps shuffled.

| input | peak RSS | wall |
|---|---|---|
| 500k rows / 223 MB | 864 MB → 371 MB (−57%) | 1.37s → 0.91s |
| 2M rows / 886 MB | 3.36 GB → 1.43 GB (−58%) | 5.77s → 3.57s |

## Verification

Against the current implementation on the same inputs: identical row count, `Timestamp` still non-decreasing, and an identical multiset of rows once `EvtxFile` is ignored. A build with changes 1 and 2 only is byte-identical, which isolates change 3 as the only source of output difference. Edge cases (header-only, no `EvtxFile` column, empty timestamp, ragged rows), `-d`, the header-mismatch warning, stdout and the `-C` guard are all unchanged. `cargo test --lib` (518) and `cargo clippy --all-targets -- -D warnings` are clean.

Four tests added, covering the per-run reset of `seen`, resetting on the sort key rather than the timestamp string (rows written in different timezones share a sort key but not a string), the input-order rule above, and the lexical fallback path.

## Notes

No CHANGELOG entry: `sort-csv` is still unreleased. Memory stays linear in the input; a timeline larger than RAM would need an external merge sort, which is out of scope. Pre-existing and untouched: the reader is `flexible(true)` but the writer is not, so ragged input fails mid-write — identical before and after this PR.

### DIFF
--- a/src/timeline/sort_csv.rs
+++ b/src/timeline/sort_csv.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process;
 
 use chrono::{DateTime, NaiveDate, NaiveDateTime, TimeZone};
-use csv::{QuoteStyle, ReaderBuilder, WriterBuilder};
+use csv::{QuoteStyle, ReaderBuilder, StringRecord, WriterBuilder};
 use hashbrown::HashSet;
 
 use crate::detections::configs::SortCsvOption;
@@ -98,7 +98,11 @@ fn collect_input_files(opt: &SortCsvOption) -> Vec<PathBuf> {
 
 /// Read one CSV, returning its header row and its data rows. The header is checked against the
 /// first file's header by the caller so that rows from different files line up column-for-column.
-fn read_csv(path: &Path) -> Option<(Vec<String>, Vec<Vec<String>>)> {
+///
+/// Rows are kept as the `csv::StringRecord` the reader already produced (2 allocations per row)
+/// instead of being converted to `Vec<String>` (one allocation per cell), which matters because
+/// every row of every input file is held in memory at once.
+fn read_csv(path: &Path) -> Option<(Vec<String>, Vec<StringRecord>)> {
     let mut reader = match ReaderBuilder::new().flexible(true).from_path(path) {
         Ok(reader) => reader,
         Err(err) => {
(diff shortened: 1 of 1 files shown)


YamatoSecurity (approved):
@fukusuket LGTM! Thanks so much!!

